Issue #2, opened 2020-07-30
Resource: trigger


kvrhdn:
A resource to create triggers.

More information: https://docs.honeycomb.io/working-with-your-data/triggers/
API: https://docs.honeycomb.io/api/triggers/

Possible Terraform config (this is not final + probably not legal Terraform code):

```hcl
resource "honeycombio_trigger" "trigger" {
  name        = "Requests are slower than usual"
  description = "Average duration of all requests for the last 5 minutes."

  # NOTE: maybe `enabled` would be simpler to grok?
  disabled = false

  # NOTE: should we prefer using terminology from the UI? I.e. `visualize`, `where`, `group by`
  query {
    # exactly one calculation is required
    calculation {
      op     = "AVG"
      column = "duration_ms"
    }

    # zero or more filter blocks
    filter {
      column = "trace.parent_id"
      op     = "does-not-exist"
    }

    # should also be supported: filter_combination, breakdowns
  }

  frequency = 300 // in seconds, 5 minutes

  threshold {
    op    = ">"
    value = 1000
  }

  # zero or more recipients
  recipient {
    type   = "email"
    target = "hello@example.com"
  }

  recipient {
    type = "pagerduty"
  }
}
```

kvrhdn:
Trigger API is now implemented in `go-honeycombio` since [v0.0.5](https://github.com/kvrhdn/go-honeycombio/releases/tag/v0.0.5).

fitzoh:
I like the  DSL but I might consider separating it from the resource.

I'm thinking of something along the lines of how AWS IAM policies are managed with the [aws_iam_policy_document data source](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document).

```hcl
data "honeycombio_query" "my_query" {
  calculation {
    ...
  }
  filter {
    ...
  }
}
resource "honeycombio_trigger" "my_trigger" {
    query = data.honeycombio_query.my_query.rendered
}
```

kvrhdn:
Oh, thanks for the link! I didn't know that was possible. This would be really nice, especially since this would allow us to share `honeycombio_query` between triggers and boards.

kvrhdn:
This is now implemented with #10.

Example code: https://github.com/kvrhdn/terraform-provider-honeycombio/blob/main/example/trigger/main.tf

PR #10, opened 2020-08-02
Add honeycombio_trigger

kvrhdn:
See #2 

Still TODO:
- [ ] add extra test case modifying the trigger
- [x] check issue with diffs (the entire block is replaced instead of just the fields)
- [x] split off the query block as datasource `honeycombio_query`
- [x] update documentation
- [ ] can we validate `query_json` has exactly one calculation?


honeycombio/query_spec.go
```diff
@@ -0,0 +1,213 @@
...
+	">=",
+	"<",
+	"<=",
+	"starts-with",
+	"does-not-start-with",
+	"exists",
+	"does-not-exist",
+	"contains",
+	"does-not-contain",
+}
```
fitzoh, line 42: Possible future TODO: Use the `go-honeycomb` enums?

https://github.com/kvrhdn/go-honeycombio/blob/eed63fb25297aadc235a18b2bef46388bd20ab7d/trigger.go#L67-L72

honeycombio/query_spec.go
```diff
@@ -0,0 +1,213 @@
...
+	return &schema.Schema{
+		Type:     schema.TypeSet,
+		Required: true,
+		MinItems: 1,
+		MaxItems: 1,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"calculation": {
+					Type:     schema.TypeSet,
+					Required: true,
```
fitzoh, line 68: Can probably make this optional:
>If no calculations are provided, COUNT is applied.

kvrhdn: You would think so, but Honeycomb actually refuses this: _request failed with status code 422: {"error":"trigger query requires exactly one calculation"}_

(You can test this by playing with `trigger_test.go` in `go-honeycombio`.)

fitzoh: Hmm...
>a strict subset of a Query Specification. A Trigger Query Specification must contain exactly one item in calculations, and may contain breakdowns, filters, and/or filter_combination, and time_range, which defines the query’s duration. The HEATMAP calculation may not be used for the trigger calculation. time_range will default to frequency if frequency is specified, or to 900 seconds (15 min, the default for frequency) if frequency is not specified

So maybe it's required for triggers, optional for other Queries (boards)?

kvrhdn: I've split off the query block as a new data source `honeycombio_query` (which makes everything much simpler!) and I've now also made `calculation` optional.
Only drawback is that creating a trigger can fail if the query does not have exactly one calculation. But maybe we can add some validation for that later.

example/trigger/main.tf
```diff
@@ -3,37 +3,36 @@ provider "honeycombio" {
...
-      op     = "="
-      value  = "ThatSpecialTenant"
-    }
-
-    # this can be ommited, AND is the default
-    filter_combination = "AND"
-
-    # also supported: breakdowns
-  }
+  query = data.honeycombio_query.query.rendered
```
fitzoh, line 35: So if we do ever decide that we want to bring the embedded DSL back as an option, it might make sense to call this `query_json`... could also be a useful hint on what the structure should look like.

fitzoh: maybe not, I could go either way on that one.
It definitely feels simpler if JSON is the only way to do it and lines up with the way `data aws_iam_policy_document` works, which is part of probably the most widely used provider out therte

honeycombio/query_spec.go
fitzoh: Looks great!

example/trigger/main.tf
kvrhdn: Good call, I'll change the field on trigger to `query_json` and the output of query to `json`. Then you get:
```hcl
data "honeycombio_query" "my_query" {
  # ...
}

resource "honeycombio_trigger" "trigger" {
  # ...

  query_json = data.honeycombio_query.my_query.json
}
```
This should make it clear how to connect the pieces.

I propose only supporting `query_json` (at least for now). Using the `honeycombio_query` should be fairly straight-forward and it's great for sharing queries between triggers and boards.
I'm wary about re-adding the nested DSL since that was fairly complicated and a lot of code to maintain.

kvrhdn:
I'm planning to merge this PR and start separate smaller PR's for the remaining TODO's. @fitzoh do you still want to review/check this PR?

fitzoh:
>I'm planning to merge this PR and start separate smaller PR's for the remaining TODO's. @fitzoh do you still want to review/check this PR?

I'd say ship it.

I'm guessing you're not too worried about API stability at this point?

Can always go back and make some tweaks later

honeycombio/query_spec.go
kvrhdn: I'm not super familiar with enums in Go. What would be the way to go? Terraform expects a `[]string`, so I guess I could create a list with all enums and map them with a utility func?
Or maybe I could add a function `isValidFilterOp(string) bool` to go-honeycombio and use that in the validator?

Issue #27, opened 2020-08-16
Query values converted to strings


fitzoh:
Creating a board using this `honeycombio_query`:

```terraform
data "honeycombio_query" "db_heatmap" {
  calculation {
    op     = "HEATMAP"
    column = "db.duration"
  }
  filter {
    column = "db.duration"
    op     = ">"
    value  = 500
  }
}
```

ends up yielding this query:
![Annotation 2020-08-15 232920](https://user-images.githubusercontent.com/131809/90325883-2e467000-df4f-11ea-97e5-d49df7aa4df7.png)

(Note that the duration is quoted)

This results in no data being shown, while data is returned if the value is converted to a numeric value.

kvrhdn:
The provider casts filter.value to a string, before assigning it: https://github.com/kvrhdn/terraform-provider-honeycombio/blob/main/honeycombio/data_source_query.go#L173
In the schema it's also declared as `TypeString`.

But in go-honeycombio filter value is a `interface{}`: https://github.com/kvrhdn/go-honeycombio/blob/main/query_spec.go#L64

So, we'll have to figure out whether it's possible to have a dynamic type in Terraform.

fitzoh:
Does it make sense to add mutually exclusive `value_string` and `value_numeric` fields?

Feels kind of gross, but probably the most straightforward option.

kvrhdn:
Yeah, I'd like to avoid having to do that :confused: but it might be the easiest way...
I'll have to investigate the Terraform type system a bit more first.

fitzoh:
From a quick look around it doesn't look like there's a way to specify that a field is a string or numeric.

My other thought was we could maybe have the go client convert it based on the operator.
This would probably work for comparison operators, but would fail on `=`/`!=` (those could be used with either string or numeric values).

The best option would probably be on the honeycomb side (we always send a string, honeycomb converts to an appropriate value for the field), but I *think* explicit value/numeric fields might be the best option for now.

But yeah, if you find a better option in the terraform type system that works too.

kvrhdn:
After playing a bit with the UI, it seems that the UI interprets the filter value based upon the type of the column. There are four types (string, integer, float, boolean) which can be set in the dataset settings.

For example, the result of entering `0.5` with the various types:

type of the column | what you type | what the UI shows after pressing enter
--|--|--
string | `0.5` | `"0.5"`
integer | `0.5` | `0`
float | `0.5` | `0.5`
[...]

PR #29, opened 2020-08-20
Add filter value properties with explicit types.

fitzoh:
As described in https://github.com/kvrhdn/terraform-provider-honeycombio/issues/27, the Honeycomb API currently fails to parse queries if a filter is sent with the wrong type in the JSON payload.

This PR adds new `value_*` options to the filter schema for each type in the Honeycomb API.
These properties are mutually exclusive with themselves and the original `value` field.

Feel free to either comment or edit @kvrhdn.
If you'd prefer to go with one of the other options enumerated in #27 that's fine too, just putting an option out there.

Fixes https://github.com/kvrhdn/terraform-provider-honeycombio/issues/27

### DIFF
--- a/honeycombio/data_source_query.go
+++ b/honeycombio/data_source_query.go
@@ -3,6 +3,7 @@ package honeycombio
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"strconv"
 
 	"github.com/hashicorp/go-cty/cty"
@@ -85,8 +86,29 @@ func dataSourceHoneycombioQuery() *schema.Resource {
 							ValidateFunc: validation.StringInSlice(validQueryFilterOps, false),
 						},
 						"value": {
-							Type:     schema.TypeString,
-							Optional: true,
+							Type:        schema.TypeString,
+							Description: "The value used for the filter.  Use of the explicitly typed `value_*` variants is recommended until the honeycomb API is able to support type inference as initially described in https://github.com/kvrhdn/terraform-provider-honeycombio/issues/27.",
+							Optional:    true,
+						},
+						"value_string": {
+							Type:        schema.TypeString,
+							Description: "The value used for the filter when the column is a string. Mutually exclusive with `value` and the other `value_*` options",
+							Optional:    true,
+						},
+						"value_integer": {
+							Type:        schema.TypeInt,
+							Description: "The value used for the filter when the column is an integer. Mutually exclusive with `value` and the other `value_*` options",
+							Optional:    true,
+						},
+						"value_float": {
+							Type:        schema.TypeFloat,
+							Description: "The value used for the filter when the column is a float. Mutually exclusive with `value` and the other `value_*` options",
+							Optional:    true,
+						},
+						"value_boolean": {
+							Type:        schema.TypeBool,
+							Description: "The value used for the filter when the column is a boolean. Mutually exclusive with `value` and the other `value_*` options",
+							Optional:    true,
 						},
 					},
 				},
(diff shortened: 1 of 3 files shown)


honeycombio/data_source_query.go
```diff
@@ -85,8 +86,29 @@ func dataSourceHoneycombioQuery() *schema.Resource {
 							ValidateFunc: validation.StringInSlice(validQueryFilterOps, false),
 						},
 						"value": {
-							Type:     schema.TypeString,
-							Optional: true,
+							Type:        schema.TypeString,
```
fitzoh, line 89: Note that having both `value` and `value_string` is redundant.
It would probably make more sense to drop `value`, but that would be a breaking change.

Also `value` will likely be the primary option once the honeycomb API is updated to do type mapping.

honeycombio/data_source_query.go
```diff
@@ -284,3 +284,71 @@ func validateQueryJSON(validators ...querySpecValidateDiagFunc) schema.SchemaVal
...
+	filters := make([]honeycombio.FilterSpec, filterSet.Len())
+	for i, schemaFilter := range filterSet.List() {
+		honeyFilter, err := extractFilter(d, hashFn(schemaFilter))
+		if err != nil {
+			return nil, err
+		}
+		filters[i] = honeyFilter
+	}
+	return filters, nil
+}
```
fitzoh, line 300: It took me a little while to figure out how to reference nested properties.
Since we're using a set instead of a list you need to find the generated hash, then you can access nested objects/properties via `d.Get('filter.<id>')`/`d.Get('filter.<id>.value')`

honeycombio/data_source_query.go
```diff
@@ -164,31 +186,9 @@ func dataSourceHoneycombioQueryRead(ctx context.Context, d *schema.ResourceData,
...
-		filter := filters[i]
-		if filter.Op == honeycombio.FilterOpExists || filter.Op == honeycombio.FilterOpDoesNotExist {
-			if filter.Value != "" {
-				return diag.Errorf("Filter operation %s must not contain a value", filter.Op)
-			}
-			filters[i].Value = nil
-		}
+	filters, err := extractFilters(d)
+	if err != nil {
+		return diag.FromErr(err)
```
fitzoh, line 191: this was getting a little unwieldy, felt like it was time to pull it into a function.

Not sure if it makes sense to do it like this or return diags all the way up

honeycombio/data_source_query.go
```diff
@@ -284,3 +284,71 @@ func validateQueryJSON(validators ...querySpecValidateDiagFunc) schema.SchemaVal
...
+			return nil, err
+		}
+		filters[i] = honeyFilter
+	}
+	return filters, nil
+}
+
+const multipleValuesError = "must choose one of 'value', 'value_string', 'value_integer', 'value_float', 'value_boolean'"
+
+func extractFilter(d *schema.ResourceData, id int) (honeycombio.FilterSpec, error) {
```
fitzoh, line 304: add a comment on the method/maybe explain the mutually exclusive value stuff

honeycombio/data_source_query.go
```diff
@@ -284,3 +284,71 @@ func validateQueryJSON(validators ...querySpecValidateDiagFunc) schema.SchemaVal
...
+	if vbOk {
+		if valueSet {
+			return filter, fmt.Errorf(multipleValuesError)
+		}
+		filter.Value = vb
+		valueSet = true
+	}
+
+	if filter.Op == honeycombio.FilterOpExists || filter.Op == honeycombio.FilterOpDoesNotExist {
+		if filter.Value != nil {
```
fitzoh, line 349: This actually became a little nicer, we can do an actual nil check instead of just seeing if it matches the empty value.

honeycombio/data_source_query.go
kvrhdn: Agreed, good call :+1:

honeycombio/data_source_query.go
kvrhdn: To investigate: the [`Schema`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2@v2.0.1/helper/schema?tab=doc#Schema) type of the Terraform SDK has a number of properties that might be able to do this validation for us:

```
	// ConflictsWith is a set of schema keys that conflict with this schema.
	// This will only check that they're set in the _config_. This will not
	// raise an error for a malfunctioning resource that sets a conflicting
	// key.
	//
	// ExactlyOneOf is a set of schema keys that, when set, only one of the
	// keys in that list can be specified. It will error if none are
	// specified as well.
	//
	// AtLeastOneOf is a set of schema keys that, when set, at least one of
	// the keys in that list must be specified.
	//
	// RequiredWith is a set of schema keys that must be set simultaneously.
	ConflictsWith []string
	ExactlyOneOf  []string
	AtLeastOneOf  []string
	RequiredWith  []string
```

I haven't used these before, so it's not clear to me how they work and if they will fit our use case.

honeycombio/data_source_query.go
kvrhdn: Indeed, this is much better 🙂

honeycombio/data_source_query.go
kvrhdn: Agree we best keep `value` around as well.

kvrhdn (commented):
Thanks for contributing this @fitzoh!. I'll check it out in more detail this evening / tomorrow.

honeycombio/data_source_query.go
fitzoh: Good catch, but it looks like it isn't supported for sets: https://github.com/hashicorp/terraform-plugin-sdk/issues/71

Tried the following:
 
`ConflictsWith: []string{"value_string", "value_integer", "value_float", "value_boolean"},`
->
```
      	* data source honeycombio_query: ConflictsWith: value references unknown
        attribute (value_string) at part (value_string)
```

---

`ConflictsWith: []string{"filter.0.value_string", "filter.0.value_integer", "filter.0.value_float", "filter.0.value_boolean"},`
->
```
        	* data source honeycombio_query: ConflictsWith: value_float configuration
        block reference (filter.0.value) can only be used with TypeList and MaxItems:
        1 configuration blocks
```

---

`ConflictsWith: []string{".value_string", ".value_integer", ".value_float", ".value_boolean"},`
->
```
        	* data source honeycombio_query: ConflictsWith: value_string references
        unknown attribute (.value) at part ()
```

PR #44, opened 2021-01-25
Introduce new resources

kvrhdn:
Depends on https://github.com/kvrhdn/go-honeycombio/pull/12

**Blocked**: the new APIs are currently in early access and deemed unstable, merge when released in beta.

---

FEATURES:

- New resource: `honeycombio_dataset`
- New data source: `honeycombio_datasets`
- New resource: `honeycombio_column`
- New resource: `honeycombio_derived_column`


docs/data-sources/datasets.md
```diff
@@ -0,0 +1,29 @@
...
+```hcl
+# returns all datasets
+data "honeycombio_datasets" "my_datasets" {
+}
+
+# only return the datasets starting with 'foo_'
+data "honeycombio_datasets" "foo_datasets" {
+  starts_with = "foo_"
+}
+```
```
fitzoh, line 16: Convenient.

If anyone from HC is taking API feedback here, it'd be nice if there was some kind of tagging functionality for datasets.

For instance if I want to add an alert to all the `prod` `web` datasets (which you could probably do via convention with prefixes)

kvrhdn: Agree this would be useful! I've copied your feedback to the Pollinators Slack 🙂

PR #60, opened 2021-04-17
Better columns, support board query annotations

### LINKED ISSUE #59: Graceful handling of 409 Conflict when creating duplicate column
Hey,

Would it be possible to have the provider not treat HTTP 409s as an error? When I try to run terraform apply with an empty state and a dataset with columns already exists, I get this error:  

```
╷
│ Error: 409 Conflict: Can't insert, column already exists with this key_name or alias.
│
│   on columns.tf line 15, in resource "honeycombio_column" "column":
│   15: resource "honeycombio_column" "column" {
│
╵
```

It'd be awesome if in this case it updated the existing column instead of having to manually import all of them.

Thanks!

tyler-boyd:
Depends on https://github.com/kvrhdn/go-honeycombio/pull/16

* Minor documentation fix
* Fixes column read/updating (it wasn't saving the ID)
* Checks if column exists before creating (fixes #59)
* Fixes query filter `value_boolean` so you can specify false and it won't ignore it

### DIFF
--- a/honeycombio/data_source_query.go
+++ b/honeycombio/data_source_query.go
@@ -272,7 +272,7 @@ func extractFilter(d *schema.ResourceData, index int) (honeycombio.FilterSpec, e
 		filter.Value = vf
 		valueSet = true
 	}
-	vb, vbOk := d.GetOk(fmt.Sprintf("filter.%d.value_boolean", index))
+	vb, vbOk := d.GetOkExists(fmt.Sprintf("filter.%d.value_boolean", index))
 	if vbOk {
 		if valueSet {
 			return filter, fmt.Errorf(multipleValuesError)
--- a/honeycombio/resource_column.go
+++ b/honeycombio/resource_column.go
@@ -71,7 +71,14 @@ func resourceColumnCreate(ctx context.Context, d *schema.ResourceData, meta inte
 	dataset := d.Get("dataset").(string)
 	column := readColumn(d)
 
-	column, err := client.Columns.Create(ctx, dataset, column)
+	existing, err := client.Columns.GetByKeyName(ctx, dataset, column.KeyName)
+
+	if err == nil {
+		d.SetId(existing.ID)
+		return resourceColumnUpdate(ctx, d, meta)
+	}
+
+	column, err = client.Columns.Create(ctx, dataset, column)
 	if err != nil {
 		return diag.FromErr(err)
 	}
@@ -119,6 +126,7 @@ func resourceColumnUpdate(ctx context.Context, d *schema.ResourceData, meta inte
 
 func readColumn(d *schema.ResourceData) *honeycombio.Column {
 	return &honeycombio.Column{
+		ID:          d.Id(),
 		KeyName:     d.Get("key_name").(string),
 		Hidden:      honeycombio.BoolPtr(d.Get("hidden").(bool)),
 		Description: d.Get("description").(string),
--- a/honeycombio/resource_derived_column.go
+++ b/honeycombio/resource_derived_column.go
@@ -66,7 +66,13 @@ func resourceDerivedColumnCreate(ctx context.Context, d *schema.ResourceData, me
 	dataset := d.Get("dataset").(string)
 	derivedColumn := readDerivedColumn(d)
 
-	derivedColumn, err := client.DerivedColumns.Create(ctx, dataset, derivedColumn)
+	existing, err := client.DerivedColumns.GetByAlias(ctx, dataset, derivedColumn.Alias)
+	if err == nil {
+		d.SetId(existing.ID)
+		return resourceDerivedColumnUpdate(ctx, d, meta)
+	}
+
+	derivedColumn, err = client.DerivedColumns.Create(ctx, dataset, derivedColumn)
 	if err != nil {
 		return diag.FromErr(err)
 	}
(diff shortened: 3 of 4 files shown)


honeycombio/resource_column.go
```diff
@@ -119,6 +133,7 @@ func resourceColumnUpdate(ctx context.Context, d *schema.ResourceData, meta inte
 
 func readColumn(d *schema.ResourceData) *honeycombio.Column {
 	return &honeycombio.Column{
+		ID:          d.Id(),
```
kvrhdn, line 136: I am wondering why this wasn't set already, was updates not working then? 🤔 Anyway, this is a good addition!

honeycombio/resource_derived_column.go
```diff
@@ -55,6 +55,13 @@ func resourceDerivedColumnImport(ctx context.Context, d *schema.ResourceData, i
 	dataset := strings.Join(idSegments[0:len(idSegments)-1], "/")
 	alias := idSegments[len(idSegments)-1]
 
+	client := meta.(*honeycombio.Client)
+	derivedColumn, err := client.DerivedColumns.GetByAlias(ctx, dataset, alias)
+	if err != nil {
+		return nil, err
+	}
+
+	d.SetId(derivedColumn.ID)
```
kvrhdn, line 64: Same here, we don't need to get and store the ID because it is set by `resourceDerivedColumnRead`.

honeycombio/resource_derived_column.go
```diff
@@ -66,7 +73,13 @@ func resourceDerivedColumnCreate(ctx context.Context, d *schema.ResourceData, me
...
 	derivedColumn := readDerivedColumn(d)
 
-	derivedColumn, err := client.DerivedColumns.Create(ctx, dataset, derivedColumn)
+	existing, err := client.DerivedColumns.GetByAlias(ctx, dataset, derivedColumn.Alias)
+	if err == nil {
+		d.SetId(existing.ID)
+		return resourceDerivedColumnUpdate(ctx, d, meta)
+	}
+
+	derivedColumn, err = client.DerivedColumns.Create(ctx, dataset, derivedColumn)
```
kvrhdn, line 82: Same here, we don't need to get and store the ID because it is set by `resourceDerivedColumnRead`.

honeycombio/resource_query_annotation.go
```diff
@@ -0,0 +1,106 @@
...
+			},
+			"description": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			// "query_id": {
+			// 	Type:     schema.TypeString,
+			// 	Required: true,
+			// 	ForceNew: true,
+			// },
```
kvrhdn, line 32: Why is this commented out? This looks good to me 🙂

honeycombio/data_source_query.go
```diff
@@ -272,7 +272,7 @@ func extractFilter(d *schema.ResourceData, index int) (honeycombio.FilterSpec, e
 		filter.Value = vf
 		valueSet = true
 	}
-	vb, vbOk := d.GetOk(fmt.Sprintf("filter.%d.value_boolean", index))
+	vb, vbOk := d.GetOkExists(fmt.Sprintf("filter.%d.value_boolean", index))
```
kvrhdn, line 275: Can you clarify this change? `GetOkExists` is marked as deprecated.

kvrhdn (changes requested):
Hi Tyler, thanks for the PR! Overall this PR contains _a lot_ of changes. I'm generally a fan of smaller PRs, that way it's easy to review and quick to merge.
This PR contains a mix of small fixes and new features. If the small fixes are submitted in a separate PR I can almost immediately merge and release them!

I'd propose splitting this PR into:

1. the small documentation fix
1. ID fix related to column and derived column
1. new query annotation resource (the new resource requires tests as well, take a look at the other resources for the general format of the acceptance tests)
1. update the board resource to support query annotation (depends on 3.)

Anyway, I've added a couple of comments 🙂

honeycombio/data_source_query.go
tyler-boyd: So `GetOk` is unable to differentiate between `value_boolean = false` and `value_boolean` being omitted completely. Not to say that there isn't a better solution than using a deprecated function, but the current usage of `GetOk` doesn't work with any filter with `value_boolean = false`

tyler-boyd:
Thanks for the feedback! I will update the PR in the next few days.

I'm OK with either making query_annotations a separate resource or just adding the ID to the tfstate so whenever we create a new one we can delete the old one. A separate resource is closer aligned to the API but considerably more verbose to use. I suppose it is better to err on the side of matching the API.

honeycombio/resource_column.go
tyler-boyd: Correct! Updates were a little broke 😅

tyler-boyd (commented):
Updated this PR to just contain the bugfixes, will open another for query annotations!

honeycombio/data_source_query.go
kvrhdn: Huh okay, I get it now. I find the implementation of `GetOk` a bit weird then tbh...
IMO It seems fine to use a deprecated function for now, eventually I'd like to refactor / improve all the `value_<type>` fields anyway.

honeycombio/resource_column.go
kvrhdn: Well I guess I didn't properly test that 😅

kvrhdn (approved):

kvrhdn:
Thanks for working on this @tyler-boyd! Hope this fixes your issues using columns and queries.

PR #61, opened 2021-04-24
Formatting markdown documents to be less than 120 characters wide

mveitas:
Update markdown documents to avoid long lines and scrolling making editing easier.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@
 terraform-provider-honeycombio*
 
 coverage.txt
+
+.idea
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,9 @@
 
 All contributions are welcome, whether they are technical in nature or not.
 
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -8,3 +8,5 @@
 terraform-provider-honeycombio*
 
 coverage.txt
+
+.idea
```
kvrhdn, line 12: Updating the .gitignore to handle every person's personal setup will lead to a lot of entries here, can I propose using a global gitignore to handle your personal preferences?

Relevant resource: https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

kvrhdn (changes requested):
I am sympathetic to the fact that limiting the line width of markdown files would be an improvement for your setup. In my own setup I've always enabled word wrap for markdown files so this isn't something I needed (this is something you can [configure globally in IntelliJ btw](https://stackoverflow.com/a/58615683/10467175)).

My main issue with formatting documents to a strict character limit is that future commits will be very noisy. When you change a paragraph by adding or removing words you will either have to deviate from the character limit or you will have to change multiple lines due to words moving between lines (= a noisy commit).

If we are going to change the formatting of all the markdown documents, I'd prefer to use something like [Semantic Line Breaks](https://sembr.org/) which will make the document easier to read and also avoid noisy commits.

mveitas:
Fair points and good feedback and something I have always done in the past. I'll close this out and look to do something in the future with your feedback in mind

PR #63, opened 2021-04-27
Adds Query resource

tyler-boyd:


tyler-boyd:
Query annotations will look like this: https://github.com/tyler-boyd/terraform-provider-honeycombio/pull/1

Very similar, so if you have any feedback I'll probably apply it to both. Will open 2nd PR after this one is merged so you don't see a messier diff.

kvrhdn:
I think this looks really good! I appreciate you updating the documentation as well :+1:

Accepting the query as `query_json` so we can reuse `data.honeycombio_query` is a good way to avoid duplicate code. Now I'm just wondering whether we should rename the `honeycombio_query` datasource to `honeycombio_query_spec` since this aligns better with the docs... 🤔 But that's whole other discussion that is not super relevant to this PR.

To be honest, it's fine for me to also add the query annotation to this PR since the files won't overlap a lot. This PR looks ready to merge, I just want to take some time to test the new resource before merging it.

tyler-boyd:
Thanks! I've merged the query-annotation branch into this one so we should see both now in this PR. I suppose this way you only need to test 1 branch, even if it's double the new resources 😄 I will have another PR later for the boards changes, it should be very small however.

👍 on renaming the datasource, I think that's a good idea.

grimesjm:
Anything we can do to help this along? :)

kvrhdn:
Not really, I just haven't had the energy to look into this :( I'll try to review (and possibly merge) this tomorrow!

kvrhdn:
CI fails because GitHub Actions does not share secrets with forks, but I can confirm the checks pass.

kvrhdn:
Sorry for really slow response, but this is great work @tyler-boyd 😃

tyler-boyd:
Thanks @kvrhdn! Hope the new job is treating you well. There will be a final small PR to link these ones together with boards.

kvrhdn:
Okay, that's great. Looking forward to it!

kvrhdn:
Hi, sorry I haven't been keeping with Honeycomb and this Terraform provider since I joined Grafana Labs. I see this PR is also pending: https://github.com/kvrhdn/terraform-provider-honeycombio/pull/67
I'll try to it merged and aim for a new release by next week.

leonsp-ai:
@kvrhdn Thanks for the great provider! I'm looking forward to the next release

Issue #66, opened 2021-06-17
Acceptance test fail due to hardcoded dataset


sleeping-barber:
Hej,
following the CONTRIBUTION.md it is open to choose a dataset name which is unfortunately not working due to this line https://github.com/kvrhdn/terraform-provider-honeycombio/blob/0cc49010a4f1337c5e69bcc53614485ef457bb36/honeycombio/resource_dataset_test.go#L32

Adjusting the description should be enough?

PR #67, opened 2021-06-17
Chore/rename data source query

### LINKED ISSUE #65: Rename data source honeycombio_query to honeycombio_query_spec
With the new honeycombio_query resource, it becomes confusing to have to differentiate between the `honeycombio_query` [data source](https://github.com/kvrhdn/terraform-provider-honeycombio/blob/main/docs/data-sources/query.md) and [the resource](https://github.com/kvrhdn/terraform-provider-honeycombio/blob/main/docs/resources/query.md).
The `honeycombio_query` data source is not actually a query though, it's a query specification as used described [in the docs](https://docs.honeycomb.io/api/query-specification/).

To avoid this confusion we should:
- rename the `honeycombio_query` data source to `honeycombio_query_spec`
- rename attributes that expect a query specification from `query` to `query_spec`

thundering-herd:
Hej @kvrhdn 
this PR should fix #65 

I renamed the data source `honeycombio_query` to `honeycombio_query_spec` and updated the documentation accordingly.

All test are passing expect one due to the issue #66 

Cheers!


docs/resources/board.md
```diff
@@ -63,7 +63,7 @@ The following arguments are supported:
 
 Each board configuration may have zero or more `query` blocks, which accepts the following arguments:
 
-* `query_json` - (Required) A JSON object describing the query according to the [Query Specification](https://docs.honeycomb.io/api/query-specification/#fields-on-a-query-specification). While the JSON can be constructed manually, it is easiest to use the [`honeycombio_query`](../data-sources/query.md) data source.
+* `query_json` - (Required) A JSON object describing the query according to the [Query Specification](https://docs.honeycomb.io/api/query-specification/#fields-on-a-query-specification). While the JSON can be constructed manually, it is easiest to use the [`honeycombio_query`](terraform-provider-honeycombio/docs/data-sources/query_spec.md) data source.
```
kvrhdn, line 66: I'm worried changing this link will break the registry docs: https://registry.terraform.io/providers/kvrhdn/honeycombio/latest/docs/resources/board
```suggestion
* `query_json` - (Required) A JSON object describing the query according to the [Query Specification](https://docs.honeycomb.io/api/query-specification/#fields-on-a-query-specification). While the JSON can be constructed manually, it is easiest to use the [`honeycombio_query`](../data-sources/query_spec.md) data source.
```

docs/resources/query.md
```diff
@@ -35,7 +35,7 @@ resource "honeycombio_query" "test_query" {
 The following arguments are supported:
 
 * `dataset` - (Required) The dataset this query is added to.
-* `query_json` - (Required) A JSON object describing the query according to the [Query Specification](https://docs.honeycomb.io/api/query-specification/#fields-on-a-query-specification). While the JSON can be constructed manually, it is easiest to use the [`honeycombio_query`](../data-sources/query.md) data source.
+* `query_json` - (Required) A JSON object describing the query according to the [Query Specification](https://docs.honeycomb.io/api/query-specification/#fields-on-a-query-specification). While the JSON can be constructed manually, it is easiest to use the [`honeycombio_query`](terraform-provider-honeycombio/docs/data-sources/query_spec.md) data source.
```
kvrhdn, line 38: Same here, this might break the registry docs
```suggestion
* `query_json` - (Required) A JSON object describing the query according to the [Query Specification](https://docs.honeycomb.io/api/query-specification/#fields-on-a-query-specification). While the JSON can be constructed manually, it is easiest to use the [`honeycombio_query`](../data-sources/query_spec.md) data source.
```

docs/resources/trigger.md
```diff
@@ -54,7 +54,7 @@ The following arguments are supported:
 
 * `name` - (Required) Name of the trigger.
 * `dataset` - (Required) The dataset this trigger is associated with.
-* `query_json` - (Required) A JSON object describng the query according to the [Query Specification](https://docs.honeycomb.io/api/query-specification/#fields-on-a-query-specification). While the JSON can be constructed manually, it is easiest to use the [`honeycombio_query`](../data-sources/query.md) data source.
+* `query` - (Required) A JSON object describng the query according to the [Query Specification](https://docs.honeycomb.io/api/query-specification/#fields-on-a-query-specification). While the JSON can be constructed manually, it is easiest to use the [`honeycombio_query`](terraform-provider-honeycombio/docs/data-sources/query_spec.md) data source.
```
kvrhdn, line 57: Same here, this might break the registry docs
```suggestion
* `query` - (Required) A JSON object describng the query according to the [Query Specification](https://docs.honeycomb.io/api/query-specification/#fields-on-a-query-specification). While the JSON can be constructed manually, it is easiest to use the [`honeycombio_query`](../data-sources/query_spec.md) data source.
```

honeycombio/resource_trigger.go
```diff
@@ -40,7 +40,7 @@ func newTrigger() *schema.Resource {
 				Type:     schema.TypeBool,
 				Optional: true,
 			},
-			"query_json": {
+			"query": {
```
kvrhdn, line 43: Why change `honeycombio_trigger`.`query_json` to `query`? This fields expects a JSON representation of the query spec, this didn't change. Board also uses `query_json`, so this would be more consistent.

This is how it's usually used:
```
resource "honeycombio_trigger" "trigger" {
  # ...

  query_json = data.honeycombio_query.query.json
}
```

This is a breaking change, so if we do this it should be deliberate and consistent with the other resources.

kvrhdn (changes requested):
Besides my notes below, this looks good and is almost ready to merge. Thanks for this! 🙂 

You'll have to merge/rebase main to get the tests working with Terraform >v1.0.1. I have tested this locally already and tests succeed :+1:

jharley:
Thank you for this work, @thundering-herd! #98 completes this work

PR #70, opened 2021-08-12
Update Board Resource to support query annotations

tyler-boyd:
Draft for now, the docs need to be updated and I don't have time at this moment to finish that.

Breaking change, removes `query_json` from boards in favour of `query_id` and `query_annotation_id`. `query_id` should probably be required, not optional.


jharley:
Interested in helping however I can to get this moving as I'd very much like to be able to have named queries on Boards

Issue #81, opened 2021-10-11
Multiline values of `expression` break Honeycomb WebUI


vlinevych:
**My use case:**
I want to use multiline strings in `expression` field of a derived column.
Web UI is clearly not designed for this, I can apply a function only if I make it one-liner.

However terraform accepts it and I've encountered the problem when I applied.

After `terraform apply` I accessed the "Derived Columns" page on the web ui
and got the blank white screen.

**The problem**
- (Honeycomb) API accepts strings that could break the "derived columns" page.

**Expected behaviour**
- (Honeycomb) API does not accept values that could break the UI
- (provider) terraform provider removes newline characters (?)

**How to reproduce**
Terraform provider allows applying multiline values

```
resource "honeycombio_derived_column" "tmp_action" {
  alias       = "tmp_action"
  dataset     = var.dataset
  description = "SLI: successful action"
  expression  = <<EOT
  IF(
    REG_MATCH($request, `PATCH https://www.whatever.[a-z]+:443/gw/api/subscriptions/[0-9]+/bar?`),
    AND(
      OR(
        STARTS_WITH($backend_status_code, "2"), 
        STARTS_WITH($backend_status_code, "3"), 
        STARTS_WITH($backend_status_code, "4")
      ),
      LTE($backend_processing_time, 2)
    )
  )
  EOT
}
```

`terraform apply`  succeeds and consequent terraform plan shows no changes. The column actually gets created. If I query the fetch_schema endpoint, I can see the column there. The newline characters are encoded as `\n` and this, probably, breaks the web interface.

```
{
      "id": "abc1234",
      "alias": "tmp_action",
      "expression": "  IF(\n    REG_MATCH($request, `PATCH https://www.whatever.[a-z]+:443/gw/api/subscriptions/[0-9]+/bar?`),\n    AND(\n      OR(\n        STARTS_WITH($backend_status_code, \"2\"), \n        STARTS_WITH($backend_status_code, \"3\"), \n        STARTS_WITH($backend_status_code, \"4\")\n      ),\n      LTE($backend_processing_time, 2)\n    )\n  )\n",
      "description": "SLI: successful action",
      "creator": {
        "created_at": "0001-01-01T00:00:00Z",
        "updated_at": "0001-01-01T00:00:00Z",
        "email": "",
        "firstName": "",
        "lastName": "",
        "companyName": "",
        "role": 0,
        "receiveUsageEmails": false,
        "id": "xxxxx"
      },
```

[...]

PR #83, opened 2021-10-15
docs/resources/derived_column: multiline expressions

vlinevych:
Added an information regarding the multiline expressions in derived columns. 

- added an example of multiline expression
- added an advise to use `trimline()`
- fixed a broken link

Tested in [doc_preview](https://registry.terraform.io/tools/doc-preview) and spellchecked.

Related issue: #81

### DIFF
--- a/docs/resources/derived_column.md
+++ b/docs/resources/derived_column.md
@@ -16,6 +16,28 @@ resource "honeycombio_derived_column" "duration_ms_log" {
 
   dataset = var.dataset
 }
+
+resource "honeycombio_derived_column" "example_multiline_expr" { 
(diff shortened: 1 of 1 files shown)


jharley:
👋🏻 @vlinevych! Thank you for this and sorry for the delay.

I believe that we actually fixed this at the API-level as I was just able to successfully make a derived column using a well-formatted heredoc with no impact to rendering the "Dataset Settings" page.

Can you please confirm that you're seeing the same now?

jharley:
I'm going to close this out. Please let me know if it's still an issue with your testing?

PR #88, opened 2021-12-09
Integrate go-honeycombio

jharley:
This imports the current [go-honeycombio](https://github.com/kvrhdn/go-honeycombio) library (with history) directly into this project with the aim to remove the external dependency until such a time that we are ready to maintain a separate Golang API client.

Commits done today (Dec 9) are the changes required to wire the client directly into the repository: I attempted to do it with as few code changes as possible.

CI is still broken from the transfer, but this builds and non-integration tests look good.


client/marker.go
```diff
@@ -0,0 +1,108 @@
...
+	markers, err := s.List(ctx, dataset)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, m := range markers {
+		if m.ID == id {
+			return &m, nil
+		}
+	}
```
dstrelau, line 90: 🙈

client/query.go
```diff
@@ -0,0 +1,39 @@
...
+type queries struct {
+	client *Client
+}
+
+// Compile-time proof of interface implementation by type queries.
+var _ Queries = (*queries)(nil)
+
+func (s *queries) Get(ctx context.Context, dataset string, id string) (*QuerySpec, error) {
+	var q QuerySpec
+	err := s.client.performRequest(ctx, "GET", fmt.Sprintf("/1/queries/%s/%s", urlEncodeDataset(dataset), id), nil, &q)
```
dstrelau, line 31: These should probably be specific that they take in the dataset slug rather than attempting to call `urlEncodeDataset`. This will work fine … until it doesn't for some specific dataset name.

dstrelau (approved):

client/query.go
jharley: Noted!

Issue #96, opened 2022-01-19
RFC: Provider Breaking Changes


jharley:
# Overview

Honeycomb has recently taken ownership of the Terraform Provider which was originally developed and maintained by community member Koenraad Verheyden.

Leading up to the first Honeycomb-owned release of the Provider there are three breaking changes we feel need to be made:

1. The `honeycombio_query` data source will be renamed to `honeycombio_query_specification`. Initially raised in #65.
1. The `honeycombio_board` resource will no longer support inline Query JSON and instead require that you create a `honeycombio_query` resource and pass the ID in the list of queries on the Board. Initially raised in #55.
1. The `honeycombio_trigger` resource will no longer support inline Query JSON and instead require that you create a `honeycombio_query` resource and pass the ID to the Trigger resource. Also raised in #55.

As the move to the Honeycomb-maintained version of the provider will require that you update the namespace used in your `required_providers` configuration from `kvrhdn` to `honeycombio` we hope that these changes will be more easily managed.

This informal RFC was written to offer transparency around these changes and give the provider's user community of the a chance to comment on the changes.

## honeycombio_query data source change

This datasource never allowed you to fetch an existing Query object, but instead helped you form a valid Query Specification in HCL. Renaming this to `honeycombio_query_specification` seems most accurate and the change should easily be handled by a simple ‘find and replace’.

## honeycombio_board resource change

The Provider currently supports providing one or more inline queries (as Query Spec JSON `query_json`) to a Board resource to create a Board. This was created before either of the Query API or Query Annotations API existed, but both customers and the field team are looking to be able to build Boards via Terraform made up of fully annotated queries managed in code.

Inline queries are problematic because providing the query JSON to the API makes a Query object for you which is then difficult to track in Terraform as it’s a “computed sometimes” attribute of the resource. The explicit creation of a Query resource also feels more declarative and Terraform-ish.

Boards with Terraform today look something like this:

```
resource “honeycombio_board” “myboard” {
    name  = "Test board managed by Terraform"
    style = "list"

    query {
      caption       = "test query"
      dataset       = var.dataset
      query_json    = data.honeycombio_query.test.json
    }
    query {
      caption        = "my verbose query"
      query_style    = "combo"
      dataset        = var.dataset
      // Inline query
      query_json  = <<EOH
{
    "time_range": 604800,
<truncated for brevity>
}
EOH
    }
}
```

The current Board resource implementation ignores the returned `query_id` completely.

### Proposed change

[...]

PR #98, opened 2022-01-21
rename query datasource to query_specification

### LINKED ISSUE #65: Rename data source honeycombio_query to honeycombio_query_spec

jharley:
The new name more accurately reflects what the data source is used for.

Breaking change as outlined in #96
Resolves #65

### DIFF
--- a/docs/data-sources/query_specification.md
+++ b/docs/data-sources/query_specification.md
@@ -1,4 +1,4 @@
-# Data Source: honeycombio_query
+# Data Source: honeycombio_query_specification
 
 Construct a query that can be used in triggers and boards. For more information about the query specification, check out [Query Specification](https://docs.honeycomb.io/api/query-specification/).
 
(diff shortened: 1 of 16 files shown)


dstrelau (approved):

jharley:
Merged in with #100

PR #113, opened 2022-02-03
A fix in the docs about trigger resource

gsinghkular:
This a small fix in [trigger docs](https://registry.terraform.io/providers/honeycombio/honeycombio/latest/docs/resources/trigger). It should say `query_id` instead of `query_json`


jharley:
Thanks for this, @gsinghkular! Much appreciated!

This fix is already in #111 and just awaiting review

Issue #102, opened 2022-01-26
Bug: Non string values (float, integer, etc.) in filters are not valid


wchrisjohnson:
We are seeing query filters fail to validate if we need to specify anything other than a string as a comparison value: float, int, etc.

**Versions**
Latest version of the provider.

**Steps to reproduce**
Use something like the following in the query definition:
```
filter {
   column 		= honeycombio_derived_column.<derived-column-name>.alias
   op     			= ">"
   value_float  	= 0.0
}
```

**Additional context**
None required.

jharley:
I just reproduced this, and confirmed this is an issue with Terraform Plugin SDK based providers: https://discuss.hashicorp.com/t/explicit-zero-values-in-provider-schemas/18744

Looking into fixes now

jharley:
Introduced by https://github.com/honeycombio/terraform-provider-honeycombio/issues/27

jharley:
@wchrisjohnson a bit of a roundabout fix for this was just released with 0.3.0. The `value_*` options will continue to have this bug and are now deprecated in favour of the cleaner-and-functional `value` (which has been _undeprecated_).

wchrisjohnson:
@jharley I apologize for not responding before now. I'll try and test this asap - I've been working elsewhere lately.

PR #114, opened 2022-02-08
Fix filter value type parsing and depreciate value_* fields

jharley:
After some testing with curl, it seemed if we properly parse-and-coerce the HCL-read string value that this would work fine without the type-specific `value_*` filter options while solving #102 in a round about fashion (and providing a much cleaner interface and nicer DX).

This also un-depreciates the filter's `value` while depreciating all of the `value_*` fields introduced by #29.

### DIFF
--- a/honeycombio/resource_board_test.go
+++ b/honeycombio/resource_board_test.go
@@ -47,7 +47,7 @@ data "honeycombio_query_specification" "test" {
   filter {
     column = "duration_ms"
     op     = ">"
-    value  = count.index 
+    value  = count.index
   }
 }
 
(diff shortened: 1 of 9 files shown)


honeycombio/resource_board_test.go
```diff
@@ -100,6 +100,11 @@ func testAccCheckBoardExists(t *testing.T, name string) resource.TestCheckFunc {
 			return fmt.Errorf("could not find created board: %w", err)
 		}
 
+		for i := range createdBoard.Queries {
```
jharley, line 103: All changes in this file should have happened in #100.

I'm happy to move them to their own commit if you feel that's better.

dstrelau (approved):

PR #129, opened 2022-03-07
Allow provider debugging

LukeCarrier:
Allow debugging with delve or an IDE's debugger by passing the `-debug` flag on start:

https://www.terraform.io/plugin/sdkv2/debugging#debugger-based-debugging

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,15 +1,29 @@
 package main
 
 import (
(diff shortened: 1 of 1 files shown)


jharley:
Thanks for this, @LukeCarrier!

Issue #123, opened 2022-03-03
Plan is always has changed with multiple recipients


aelmekeev:
If we have a trigger with multiple recipients than we would never have an empty plan after the apply.

**Versions**
```
    honeycombio = {
      source  = "honeycombio/honeycombio"
      version = "~> 0.3.0"
    }
```

**Steps to reproduce**

```
terraform {
  required_providers {
    honeycombio = {
      source  = "honeycombio/honeycombio"
      version = "~> 0.3.0"
    }
  }
}

provider "honeycombio" {
  api_key = "<api_key_here>"
}

locals {
  dataset = "prod"
}

data "honeycombio_query_specification" "example" {
  calculation {
    op     = "AVG"
    column = "duration_ms"
  }

  filter {
    column = "trace.parent_id"
    op     = "does-not-exist"
  }
}

resource "honeycombio_query" "example" {
  dataset    = local.dataset
  query_json = data.honeycombio_query_specification.example.json
}

data "honeycombio_trigger_recipient" "slack" {
  dataset = local.dataset
  type    = "slack"
  target  = "#alex-test"
}

resource "honeycombio_trigger" "example" {
  name        = "Dummy trigger"
  description = "Dummy trigger"

  query_id = honeycombio_query.example.id
  dataset  = local.dataset

  frequency = 6000 // in seconds, 10 minutes

  threshold {
    op    = ">"
    value = 100000000
  }

  recipient {
    type   = "pagerduty"
    id = "2TLD7x6yunN"
    target = null
  }

  recipient {
    type   = "slack"
    id = data.honeycombio_trigger_recipient.slack.id
    target = "#alex-test"
  }
}
```

Run:
* `terraform apply`
* `terraform apply` again plan is not empty

Same happens if you swap recipient blocks order.

**Additional context**

The next step issues happens if you try to apply the changes since we see the following error:

```
Terraform will perform the following actions:

  # honeycombio_trigger.example will be updated in-place
  ~ resource "honeycombio_trigger" "example" {
        id          = "yF8JrPtDZ7q"
        name        = "Pagerduty dummy"
        # (5 unchanged attributes hidden)

      ~ recipient {
          ~ id     = "2TLD7x6yunN" -> "yWmLrb34bsL"
          + target = "#alex-test"
          ~ type   = "pagerduty" -> "slack"
        }
      ~ recipient {
          ~ id     = "EYNPDUi675" -> "2TLD7x6yunN"
[...]

PR #130, opened 2022-03-07
Force changes to recipients to recreate the trigger

LukeCarrier:
This is a(n ugly) workaround for #123 -- instead of attempting to update recipients in-place, which can lead to issues where the order of recipients as returned by API responses doesn't match their ordering in the Terraform configuration causes spurious or invalid diffs.

A better fix would be to make the `recipient` field `schema.TypeSet`, but [this causes issues with computed and optional values](https://github.com/honeycombio/terraform-provider-honeycombio/issues/123#issuecomment-1060630473).


LukeCarrier:
Better mitigation in #132

PR #144, opened 2022-04-07
Fix for 'AND' filter combination causing noisy plans

### LINKED ISSUE #141: Setting filter_combination to AND leads to noisy plan
**Versions**
```hcl
    honeycombio = {
      source  = "honeycombio/honeycombio"
      version = "~> 0.3.0"
    }
```

**Steps to reproduce**

1. Run `terraform apply` to create a trigger with `filter_combination` set to `AND` in `honeycombio_query_specification`.
2. Run `terraform plan`.

Expected result: plan is empty
Actual results: plan is not empty since it tries to add filter combination with `AND` again:

```hcl
  # module.test.honeycombio_query.this[0] must be replaced
-/+ resource "honeycombio_query" "this" {
      ~ id         = "JrEwdX44JCu" -> (known after apply)
      ~ query_json = jsonencode(
          ~ {
              + filter_combination = "AND"
                # (3 unchanged elements hidden)
            } # forces replacement
        )
        # (1 unchanged attribute hidden)
    }

  # module.test.honeycombio_trigger.this["#alex-test"] will be updated in-place
  ~ resource "honeycombio_trigger" "this" {
        id          = "prn53msd1kA"
        name        = "Alex test (#alex-test)"
      ~ query_id    = "JrEwdX44JCu" -> (known after apply)
        # (4 unchanged attributes hidden)


        # (2 unchanged blocks hidden)
    }
```

**Additional context**

Looks like API returns null for filter combination if it set to "default" `AND`.
We can workaround it with something like
```hcl
filter_combination = var.filter_combination == "AND" ? null : var.filter_combination
```
but it would be nice to make it more user friendly.

jharley:
Resolves #141

### DIFF
--- a/honeycombio/data_source_query_specification.go
+++ b/honeycombio/data_source_query_specification.go
@@ -111,9 +111,9 @@ func dataSourceHoneycombioQuerySpec() *schema.Resource {
 				},
 			},
 			"filter_combination": {
-				Type:     schema.TypeString,
-				Optional: true,
-				// Note: API assumes 'AND' if not provided
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "AND",
 				ValidateFunc: validation.StringInSlice([]string{"AND", "OR"}, false),
 			},
 			"breakdowns": {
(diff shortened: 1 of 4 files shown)


dstrelau:
Feels like the API should just return `AND` when it's `AND`. Let's fix that.

jharley:
I'm into that if you are! Lines 114-116 should still go in if that's the case

dstrelau:
Unfortunately, fixing this API side is a bit more complicated than I expected. Let's go with this for now.

dstrelau (approved):

PR #159, opened 2022-05-05
[triggers] Add alert_type field for Triggers

reulan:
Adding a new attribute to Triggers, it will include the following changes:
- adding a new alert frequency to the Trigger structure

The current behaviors of triggers is an alert is sent on change. We are adding a the feature for the provider set alert on true for our triggers.

Read more about alert frequency here:
https://docs.honeycomb.io/working-with-your-data/triggers/#configuration

---
  - https://app.asana.com/0/0/1202136190006122

### DIFF
--- a/client/trigger.go
+++ b/client/trigger.go
@@ -58,7 +58,7 @@ type Trigger struct {
 	// be greater than 4 times the frequency.
 	Query   *QuerySpec `json:"query,omitempty"`
 	QueryID string     `json:"query_id,omitempty"`
-	// Alert Type. contain types of trigger schedule to use. This field is required
+	// Alert Frequency. Describes scheduling behavior for triggers. By default alert type per change. This field is required
 	AlertFrequency string `json:"alert_frequency,omitempty"`
 	// Threshold. This fild is required.
 	Threshold *TriggerThreshold `json:"threshold"`
--- a/client/trigger.go
+++ b/client/trigger.go
@@ -97,14 +97,6 @@ func TriggerThresholdOps() []TriggerThresholdOp {
 	}
 }
 
-// Representation of the Triggers Alert Type.
-type TriggerAlertType struct {
-	Type TriggerAlertTypeValue `json:"alert_type"`
-}
-
-// TriggerAlertTypeValue, the current setting type for the trigger.
-type TriggerAlertTypeValue string
-
 // Allowed values for alert_type. | on_change is default
 const (
 	TriggerAlertTypeValueOnChange string = "on_change"
--- a/client/trigger.go
+++ b/client/trigger.go
@@ -59,7 +59,7 @@ type Trigger struct {
 	Query   *QuerySpec `json:"query,omitempty"`
 	QueryID string     `json:"query_id,omitempty"`
 	// Alert Frequency. Describes scheduling behavior for triggers. By default alert type per change. This field is required
-	AlertFrequency string `json:"alert_frequency,omitempty"`
+	AlertType string `json:"alert_type,omitempty"`
 	// Threshold. This fild is required.
 	Threshold *TriggerThreshold `json:"threshold"`
 	// Frequency describes how often the trigger should run. Frequency is an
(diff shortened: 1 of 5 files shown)


reulan:
need to update the local Terraform docs with the new Trigger attribute "alert_type", otherwise API changes have been tested with local version of this provider + local version of the API. Will test this again the live API once the changes are there, and then will have this is a good to go state!

honeycombio/resource_trigger.go
```diff
@@ -45,6 +45,10 @@ func newTrigger() *schema.Resource {
 				Type:     schema.TypeString,
 				Required: true,
 			},
+			"alert_type": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
```
jharley, line 51: We should have the default provided here (`on_change`) and switch this to optional otherwise this will end up being a breaking change (this will also match the documentation above :)).

Another nice user-focused improvement would be to add a `ValidateFunc` using ` validation.StringInSlice()` and checking that it is one of the supported values. Take a look at `filter_combination` in the query specification data source as an example: https://github.com/honeycombio/terraform-provider-honeycombio/blob/main/honeycombio/data_source_query_specification.go#L113-L117

honeycombio/resource_trigger_test.go
```diff
@@ -79,6 +81,20 @@ func testAccCheckTriggerAttributes(t *honeycombio.Trigger) resource.TestCheckFun
...
 
+		if t.AlertType != "on_change" {
+			return fmt.Errorf("bad AlertType: %s | should be on_change", t.AlertType)
+		}
+
+		return nil
+	}
+}
+
+func testAccCheckTriggerAlertType(t *honeycombio.Trigger) resource.TestCheckFunc {
```
jharley, line 92: By using the ValidateFunc above this won't be needed

docs/resources/trigger.md
```diff
@@ -64,6 +66,7 @@ The following arguments are supported:
 * `description` - (Optional) Description of the trigger.
 * `disabled` - (Optional) The state of the trigger. If true, the trigger will not be run. Defaults to false.
 * `frequency` - (Optional) The interval (in seconds) in which to check the results of the query’s calculation against the threshold. Value must be divisible by 60 and between 60 and 86400 (between 1 minute and 1 day). Defaults to 900 (15 minutes).
+* `alert_type` - (Optional) The frequency for the alert to trigger. (on_change is the default behavior, on_true can also be selected)
```
jharley, line 69: Nit: `on_change` and `on_true`
```suggestion
* `alert_type` - (Optional) The frequency for the alert to trigger. (`on_change` is the default behavior, `on_true` can also be selected)
```

jharley (changes requested):
Thank you for this! Lookin' pretty good! 🥳 

A few things to tidy up to offer a better UX, but nearly there

client/trigger.go
```diff
@@ -58,6 +58,8 @@ type Trigger struct {
 	// be greater than 4 times the frequency.
 	Query   *QuerySpec `json:"query,omitempty"`
 	QueryID string     `json:"query_id,omitempty"`
+	// Alert Frequency. Describes scheduling behavior for triggers. By default alert type per change. This field is required
```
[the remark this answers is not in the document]
reulan: nope not required! need to go back and clean this up

honeycombio/resource_trigger_test.go
reulan: awesome, will remove testAccCheckTriggerAlertType then

honeycombio/resource_trigger.go
```diff
@@ -45,6 +45,11 @@ func newTrigger() *schema.Resource {
 				Type:     schema.TypeString,
 				Required: true,
 			},
+			"alert_type": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "on_change",
```
jharley, line 51:
```suggestion
				Default:  "on_change",
```
```suggestion
				Default:  "on_change",
				ValidateFunc: validation.StringInSlice([]string{"on_change", "on_true"}, false),
```

client/trigger.go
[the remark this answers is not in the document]
jharley:
```suggestion
	// Alert Frequency. Describes scheduling behavior for triggers. By default alert type per change. This field is required
```
```suggestion
	// Alert Frequency. Describes scheduling behavior for triggers. By default alert type per change.
```

honeycombio/resource_trigger_test.go
```diff
@@ -208,6 +218,48 @@ resource "honeycombio_trigger" "test" {
 }`, dataset, dataset, frequency)
 }
 
+func testAccTriggerConfigWithAlertType(dataset string, alertType string) string {
```
jharley, line 221: I don't think this gets called? Cool to just rely on what you've done above and remove it

jharley, line 221: I don't think this gets called? Cool to just rely on what you've done above and remove it

jharley (changes requested):
Looking better! Few final things to clear up

honeycombio/resource_trigger_test.go
jharley: Alternatively, you can add it in as a test step above and test with both `on_change` and `on_true`

reulan: ah yep, the test is not used - this was from my initial first pass. can remove in lieu of the validation written in

jharley (approved):

jharley:
Closes #163

PR #165, opened 2022-05-16
datasource/query_result - take query JSON directly

jharley:
By taking a Query ID it was not possible to use the results of the data source without first generating an output (or similar) making use of the results.

By switching to taking Query JSON directly, Terraform is able to use the results later in the same plan as no resources need to be created first.

Note: this is a breaking change


jharley:
Just checked and no Honeycomb users (except me) have started using this data source yet so the breaking change has basically zero impact.

dstrelau (approved):

PR #166, opened 2022-05-16
Initial SLO and BurnAlert support

jharley:


mjayaram (approved):

honeycombio/type_helpers.go
```diff
@@ -97,8 +97,16 @@ func sortOrderStrings() []string {
 	return out
 }
 
-func triggerRecipientTypeStrings() []string {
-	in := honeycombio.TriggerRecipientTypes()
+func recipientTypeStrings(alertType string) []string {
```
dstrelau, line 100: Using a string param like this is kind of icky. Might be better to pass in a const or maybe `honeycombio.TriggerRecipientTypes()` directly. Alternatively, could define `type RecipientTypes []RecipientType` and put something like a `AsString() []string` method on it so this helper can be directly on the type.

honeycombio/resource_slo.go
```diff
@@ -0,0 +1,166 @@
...
+	dataset := d.Get("dataset").(string)
+
+	s, err := client.SLOs.Get(ctx, dataset, d.Id())
+	if err != nil {
+		if err == honeycombio.ErrNotFound {
+			d.SetId("")
+			return nil
+		}
+		return diag.FromErr(err)
+	}
```
dstrelau, line 114: I think this would be a bit clearer in a combined if/else if (same in burn alerts too):

```suggestion
	if err == honeycombio.ErrNotFound {
		d.SetId("")
		return nil
	} else if err != nil {
		return diag.FromErr(err)
	}
```

dstrelau (approved):

PR #171, opened 2022-05-31
SLO Docs

brookesargent:
- Add docs for `honeycombio_recipient`, `honeycombio_slo`, and `honeycombio_burn_alert`. 
- Add deprecation warning for `honeycombio_trigger_recipient`.


honeycombio/data_source_trigger_recipient.go
```diff
@@ -28,6 +28,7 @@ func dataSourceHoneycombioSlackRecipient() *schema.Resource {
 				Optional: true,
 			},
 		},
+		DeprecationMessage: "Use honeycombio_recipient data_source instead",
```
jharley, line 31: Love this! I think it is worth adding to the trigger_recipient data source as a doc note as well

docs/resources/burn_alert.md
```diff
@@ -0,0 +1,60 @@
...
+## Argument Reference
+
+The following arguments are supported:
+
+* `slo_id` - (Required) ID of the SLO this burn alert is associated with.
+* `dataset` - (Required) The dataset this burn alert is associated with.
+* `exhaustion_minutes` - (Required) The amount of time, in minutes, remaining before the SLO's error budget will be exhausted and the alert will fire.
+* `recipient` - (Optional) Zero or more configuration blocks (described below) with the recipients to notify when the alert fires.
+
+Each burn alert configuration may have zero or more `recipient` blocks, which each accept the following arguments. A recipient block can either refer to an existing recipient (a recipient that is already present in another burn alert or trigger) or a new recipient. When specifying an existing recipient, only `id` may be set. If you pass in a recipient without its ID and only include the type and target, Honeycomb will make a best effort to match to an existing recipient. To retrieve the ID of an existing recipient, refer to the [`honeycombio_recipient`](../data-sources/trigger_recipient.md) data source.
```
jharley, line 43: Links to hold data source

```suggestion
Each burn alert configuration may have zero or more `recipient` blocks, which each accept the following arguments. A recipient block can either refer to an existing recipient (a recipient that is already present in another burn alert or trigger) or a new recipient. When specifying an existing recipient, only `id` may be set. If you pass in a recipient without its ID and only include the type and target, Honeycomb will make a best effort to match to an existing recipient. To retrieve the ID of an existing recipient, refer to the [`honeycombio_recipient`](../data-sources/recipient.md) data source.
```

jharley (approved):
Minor link ref and a thought on updating the old datasource.

LGTM otherwise 🚀

PR #185, opened 2022-07-13
[datasets] Add additional fields description and ExpandJSON Depth

reulan:
Adding the following field for Honeycomb Datasets:
- `description`
-  `ExpandJSONDepth`

These changes are currently be added to the shep API seperately.

Asana Tracking:
https://app.asana.com/0/1202083175313322/1202555362164075


client/dataset.go
```diff
@@ -33,8 +37,16 @@ var _ Datasets = (*datasets)(nil)
...
 type Dataset struct {
-	Name string `json:"name"`
-	Slug string `json:"slug,omitempty"`
+	Name            string  `json:"name"`
+	Description     *string `json:"description,omitempty"`
+	Slug            string  `json:"slug,omitempty"`
+	ExpandJSONDepth *int    `json:"expand_json_depth,omitempty"`
+}
+
+type DatasetCreateArgs struct {
```
jharley, line 46: I would rather see `Dataset` follow the same update pattern as all of the other resources in the provider's client instead of this approach with the `CreateArgs` struct. It may not be ideal the way it is done, but there is something to be said for consistency in a codebase.

client/dataset_test.go
```diff
@@ -12,10 +12,14 @@ func TestDatasets(t *testing.T) {
...
 	c := newTestClient(t)
 	datasetName := testDataset(t)
+	datasetDescription := ""
+	datasetExpandJSONDepth := 0
 
 	currentDataset := &Dataset{
-		Name: datasetName,
-		Slug: urlEncodeDataset(datasetName),
+		Name:            datasetName,
+		Description:     &datasetDescription,
```
jharley, line 20: I think _not_ setting these here proves that they are truly not required and that the default values work as intended

client/dataset_test.go
```diff
@@ -47,4 +51,16 @@ func TestDatasets(t *testing.T) {
 		assert.NoError(t, err)
 		assert.Equal(t, currentDataset, d)
 	})
+
+	t.Run("Update", func(t *testing.T) {
+		updateDataset := &Dataset{
+			Name:            datasetName,
+			Description:     &datasetDescription,
```
jharley, line 58: I'd like to see one of `Description` or `ExpandJSONDepth` set to a new value to test the update

docs/resources/dataset.md
```diff
@@ -11,17 +11,20 @@ Creates a dataset.
...
+  description = "buzzing with data"
 }
 ```
 
 ## Argument Reference
 
-* `name` - (Required) The name of the dataset.
+-   `name` - (Required) The name of the dataset.
```
jharley, line 22: Not sure if this was a markdown linter, but in the name of consistency I'd like to stick to the `*` approach used everywhere else

docs/resources/dataset.md
```diff
@@ -11,17 +11,20 @@ Creates a dataset.
...
+-   `description` - Longer description for datset.
+-   `expand_json_depth` - How many JSON levels to unfurl.
 
 ## Attribute Reference
 
 In addition to all arguments above, the following attributes are exported:
 
-* `slug` - The slug of the dataset.
+-   `slug` - The slug of the dataset.
```
jharley, line 30: Same as above

honeycombio/data_source_datasets_test.go
```diff
@@ -37,6 +38,44 @@ func TestAccDataSourceHoneycombioDataset_basic(t *testing.T) {
 	})
 }
 
+func TestAccDataSourceHoneycombioDataset_createArgs(t *testing.T) {
```
jharley, line 41: I don't think this test is needed once we remove this `CreateArgs` approach

honeycombio/resource_dataset.go
```diff
@@ -23,19 +23,39 @@ func newDataset() *schema.Resource {
 				ForceNew:     true,
 				ValidateFunc: validation.StringLenBetween(1, 255),
 			},
+			"description": {
+				Type:         schema.TypeString,
```
jharley, line 27: I don't think we need to set every default here as its not done elsewhere

honeycombio/resource_dataset.go
```diff
@@ -23,19 +23,39 @@ func newDataset() *schema.Resource {
...
+				Optional:     true,
+				ForceNew:     false,
+				ValidateFunc: validation.StringLenBetween(0, 1024),
+			},
 			"slug": {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"expand_json_depth": {
+				Type:         schema.TypeInt,
```
jharley, line 39: I don't think we need to set every default here as its not done elsewhere

honeycombio/resource_dataset.go
```diff
@@ -23,19 +23,39 @@ func newDataset() *schema.Resource {
...
+			},
 		},
 	}
 }
 
 func resourceDatasetCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	client := meta.(*honeycombio.Client)
 
+	description := d.Get("description").(string)
+	expandJSONDepth := d.Get("expand_json_depth").(int)
```
jharley, line 54: In the name of consistency, I think we should move this into the struct initialization or bring `name` out of the struct initialization

honeycombio/resource_dataset.go
```diff
@@ -59,6 +79,27 @@ func resourceDatasetRead(ctx context.Context, d *schema.ResourceData, meta inter
...
 	d.Set("slug", dataset.Slug)
 	return nil
 }
+
+func resourceDatasetUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*honeycombio.Client)
+
+	description := d.Get("description").(string)
+	expandJSONDepth := d.Get("expand_json_depth").(int)
+	data := &honeycombio.Dataset{
```
jharley, line 93: As above

honeycombio/resource_dataset_test.go
```diff
@@ -8,11 +8,16 @@ import (
...
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	honeycombio "github.com/honeycombio/terraform-provider-honeycombio/client"
 	"github.com/stretchr/testify/assert"
 )
 
 func TestAccHoneycombioDataset_basic(t *testing.T) {
-	testDataset := testAccDataset()
+
+	testDatasetName := testAccDataset()
+	testDataset := honeycombio.Dataset{
```
jharley, line 18: Rather than using the test dataset here I think it makes sense to make a _new_ dataset and then validate its attributes

jharley (changes requested):
Thanks for this!

Few things to tidy up and then I think we're good to go 👍🏻

docs/resources/dataset.md
reulan: yep, this must of been a plugin of mine, was not intentional.

client/dataset_test.go
```diff
@@ -12,12 +12,23 @@ func TestDatasets(t *testing.T) {
...
 	c := newTestClient(t)
 	datasetName := testDataset(t)
+	updatedDescription := "buzzing with data"
+	updatedExpandJSONDepth := 3
 
 	currentDataset := &Dataset{
-		Name: datasetName,
-		Slug: urlEncodeDataset(datasetName),
+		Name:            datasetName,
+		Description:     &updatedDescription,
```
jharley, line 20: I think if you remove `Description` and `ExpandJSONDepth` from this `currentDataset` initialization you will get passing tests.

Because we can't delete datasets via the API today, you should be sure to have a cleanup function that resets the dataset to the default values as well

docs/resources/dataset.md
```diff
@@ -11,17 +11,20 @@ Creates a dataset.
...
+  description = "buzzing with data"
 }
 ```
 
 ## Argument Reference
 
-* `name` - (Required) The name of the dataset.
+*   `name` - (Required) The name of the dataset.
```
jharley, line 22: Nit bit the spacing is still off here

client/dataset_test.go
reulan: thanks for the comments, will be getting back to addressing these changes today

docs/resources/dataset.md
```diff
@@ -19,6 +20,8 @@ resource "honeycombio_dataset" "my_dataset" {
 The following arguments are supported:
 
 * `name` - (Required) The name of the dataset.
+* `description` - Longer description for datset.
```
jharley, line 23:
```suggestion
* `description` - (Optional) A longer description for dataset.
```

docs/resources/dataset.md
```diff
@@ -19,6 +20,8 @@ resource "honeycombio_dataset" "my_dataset" {
...
+* `expand_json_depth` - How many JSON levels to unfurl.
```
jharley, line 24:
```suggestion
* `expand_json_depth` - (Optional) The maximum unpacking depth of nested JSON fields.
```

honeycombio/resource_dataset.go
```diff
@@ -24,10 +24,22 @@ func newDataset() *schema.Resource {
 				ForceNew:     true,
 				ValidateFunc: validation.StringLenBetween(1, 255),
 			},
+			"description": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     false,
```
jharley, line 30: Nit: would still like to see `ForceNew: false` removed
```suggestion
```

honeycombio/resource_dataset.go
```diff
@@ -24,10 +24,22 @@ func newDataset() *schema.Resource {
...
+				ValidateFunc: validation.StringLenBetween(0, 1024),
+			},
 			"slug": {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"expand_json_depth": {
+				Type:         schema.TypeInt,
+				Optional:     true,
+				ForceNew:     false,
```
jharley, line 40: Nit: would still like to see ForceNew: false removed
```suggestion
```

jharley (commented):
Four quick edits then we're good to go! 🎉

jharley (approved):

Issue #117, opened 2022-02-12
Add HONEYCOMB_API_KEY envvar support


jharley:
The Honeycomb standard has become `HONEYCOMB_API_KEY` so the provider should support it too in the name of consistency.

This will deprecate but continue to support the current `HONEYCOMBIO_APIKEY` but drop support altogether for the 1.0.0 release.

teddylear:
@jharley I can take a look into this if that's alright.

jharley:
Closed by #187

PR #187, opened 2022-07-15
feat: Add support for HONEYCOMB_API_KEY environment variable for provider

teddylear:
feat: Add support for HONEYCOMB_API_KEY environment variable for provider

@jharley Converted current tests to use new environment variable and updated documentation saying that both are supported, but new env var takes priority. If there's anything I should change or test please let me know. 

Also I count not get the tests working locally with my API key, does it required an enterprise account for SLO tests? 

Addresses #117


jharley (changes requested):
Hey, @teddylear! Thanks for picking this up.

This is a great, but the intended variable name is `HONEYCOMB_API_KEY` (to match many of our other integrations) and you've got `HONEYCOMBIO_API_KEY`. If you could correct that we can get this in (and I'll add the new variable to CI)

teddylear:
Ahh Sorry I missed that, I can update.

teddylear:
@jharley Updated to use correct env var name now, please let me know if there's anything else I should test/add/change. Also, thanks for the quick initial review!

honeycombio/provider.go
```diff
@@ -25,10 +25,13 @@ func Provider() *schema.Provider {
...
-				Type:        schema.TypeString,
-				Required:    true,
-				DefaultFunc: schema.EnvDefaultFunc("HONEYCOMBIO_APIKEY", nil),
-				Sensitive:   true,
+				Type:     schema.TypeString,
+				Required: true,
+				// Tiering goes HONEYCOMB_API_KEY > HONEYCOMBIO_APIKEY > nil
+				DefaultFunc: schema.EnvDefaultFunc(
+					"HONEYCOMB_API_KEY", schema.EnvDefaultFunc(
+						"HONEYCOMBIO_APIKEY", nil)),
```
jharley, line 33: I'd like to see a call to `tflog.Warn()` here to notify people that they're using the deprecated env var. What do you think?

teddylear: I like that idea, let me add it.

teddylear: @jharley Ok added change, please let me know if there's anything else to update or change.

honeycombio/provider.go
```diff
@@ -21,14 +22,25 @@ func init() {
 // overwritten during the release process.
 var providerVersion = "dev"
 
+func deprecatedEnvCheck() schema.SchemaDefaultFunc {
+	if _, ok := os.LookupEnv("HONEYCOMBIO_APIKEY"); ok {
+		ctx := context.Background()
+		tflog.Warn(ctx, "HONEYCOMBIO_APIKEY is being depcrated, please use HONEYCOMB_API_KEY instead")
```
jharley, line 28: Minor typo/phrasing fix, but this is great! ✨ 

```suggestion
		tflog.Warn(ctx, "HONEYCOMBIO_APIKEY has been deprecated, please use HONEYCOMB_API_KEY instead")
```

teddylear: Ahhh let me make change.

jharley: Thank you! We'll get this into 0.9.0

teddylear:
@jharley Committed your suggested change, let me know if there's anything else I should change.

jharley (approved):
Lookin' good! Thanks for this, @teddylear 🎉

PR #188, opened 2022-07-15
Additional recipients work

jharley:
adds the remaining recipient resources (slack, pagerduty, and webhook) and a pair of data sources to lookup existing recipients

Relates to #186, and #8


honeycombio/resource_slack_recipient.go
```diff
@@ -0,0 +1,94 @@
...
+				Type:         schema.TypeString,
+				Required:     true,
+				Description:  "The Slack channel or username to send the notification to. Must begin with `#` or `@`.",
+				ValidateFunc: validation.StringMatch(regexp.MustCompile(`^(#|@).+`), "channel must begin with `#` or `@`"),
+			},
+		},
+	}
+}
+
+func resourceSlackRecipientCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
dstrelau, line 35: It's these CRUD functions that all look really similar to me. Can they be cleaned up at all, or is this just how it is?

honeycombio/resource_slack_recipient.go
```diff
@@ -0,0 +1,94 @@
...
+		return diag.FromErr(err)
+	}
+
+	r, err = client.Recipients.Create(ctx, r)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	d.SetId(r.ID)
+	return resourceSlackRecipientRead(ctx, d, meta)
```
dstrelau, line 49: Is this a standard pattern? We re-read the result from the API? 🤔

jharley: Yep: I think every resource in the provider works this way. I believe the aim is that with this approach the API is authoritative

dstrelau (approved):

client/recipient_test.go
```diff
@@ -57,6 +57,7 @@ func TestRecipientsEmail(t *testing.T) {
 
 		assert.NoError(t, err, "failed to update Recipient")
 		// copy dynamic field before asserting equality
+		rcpt.CreatedAt = result.CreatedAt
```
jharley, line 60: This is a hack.. the client tests need to be revisited in general as they're starting to show strain under more frequent usage.

client/trigger_test.go
```diff
@@ -51,10 +51,6 @@ func TestTriggers(t *testing.T) {
 				Value: 10000,
 			},
 			Recipients: []NotificationRecipient{
```
jharley, line 53: This is a hack.. the client tests need to be revisited in general as they're starting to show strain under more frequent usage.

honeycombio/data_source_recipient.go
```diff
@@ -13,44 +13,109 @@ func dataSourceHoneycombioRecipient() *schema.Resource {
 	return &schema.Resource{
 		ReadContext: dataSourceHoneycombioRecipientRead,
 
+		Description: `
+'honeycombio_recipient' data source provides details about a specific recipient.
+
+The ID of an existing recipient can be used when adding recipients to triggers or burn alerts.
+
+Note: If more or less than a single match is returned by the search, Terraform will fail. Ensure
```
dstrelau, line 21: Phrasing is a bit awkward here. I might prefer "Terraform will fail unless exactly one recipient is returned by the search."

honeycombio/data_source_recipients.go
```diff
@@ -0,0 +1,107 @@
...
+							ValidateFunc:  validation.StringIsValidRegExp,
+						},
+					},
+				},
+			},
+			"ids": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
```
dstrelau, line 66: Is there any reason to restrict this to only returning the IDs instead of having the whole recipient object available?

honeycombio/data_source_recipients.go
```diff
@@ -0,0 +1,107 @@
...
+		ReadContext: dataSourceHoneycombioRecipientsRead,
+
+		Description: `
+'honeycombio_recipients' data source provides recipient IDs of recipients matching a set of criteria.
+`,
+
+		Schema: map[string]*schema.Schema{
+			"type": {
+				Type:         schema.TypeString,
+				Optional:     true,
```
dstrelau, line 26: As discussed, I think `type` should be required if there is a `detail_filter` block, as the filter sort of implicitly adds the type filter anyway. Best to make it explicit.

honeycombio/data_source_recipients.go
jharley: I had initially went about ~50% of the way to return list of Recipient objects and then talked myself out of it looking at [aws_subnets](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/subnets#attributes-reference).

If you're feeling that makes sense too, I'll do that.

Each object (in JSON for lazy reasons) would be something like:

```json
{
  "id": "jdsfjksdf"
  "type": "pagerduty",
  "integration_key": "ksdfksdfkskdf",
  "integration_name": "my service"
}
```

or

```json
{
  "id": "jdsfj83f"
  "type": "email",
  "address": "alerts@example.net"
}
```

Make sense?

jharley: I think this means the attributes of the single recipient returned by the `honeycombio_recipient` data source should also populate these for consistency? 🤔

honeycombio/data_source_recipient.go
jharley: Agreed! Updated.

honeycombio/data_source_recipients.go
dstrelau: AWS data sources are sort of all over the place by the look of it. Eg, [aws_kinesis_stream](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/subnets#attributes-reference) has quite a few attributes.

I'd say there's no harm in providing more data here, but we may want to skip sensitive information like `integration_key`.

jharley: Agreed, they are all over the place. But, the "plural" versions tend to just return the IDs (often for use by the singular). For example, you can use `aws_availability_zones` to get all of the AZs matching a filter and then iterate over that list of IDs with `aws_availability_zone`. (there is no `aws_kinesis_streams` for comparison)

... maybe I'll go that route and enrich the single result from `honeycombio_recipient` with all the additional fields? 🤔

dstrelau: Ah, interesting! That sounds good 👍

jharley: Okay, I think we got there.

dstrelau (approved):

PR #191, opened 2022-07-22
Adds severity support for PagerDuty notifications

jharley:
BurnAlerts and Triggers will now be able to optionally pass a severity level to PagerDuty when a notification is sent.

---
  - https://app.asana.com/0/0/1202483433776325

### DIFF
--- a/honeycombio/type_helpers.go
+++ b/honeycombio/type_helpers.go
@@ -164,6 +164,15 @@ func flattenNotificationRecipients(rs []honeycombio.NotificationRecipient) []map
 			"type":   string(r.Type),
 			"target": r.Target,
 		}
+		if r.Details != nil {
+			// notification details have been provided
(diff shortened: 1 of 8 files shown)


honeycombio/type_helpers.go
```diff
@@ -164,6 +164,15 @@ func flattenNotificationRecipients(rs []honeycombio.NotificationRecipient) []map
 			"type":   string(r.Type),
 			"target": r.Target,
 		}
+		if r.Details != nil {
+			// notification details have been provided
+			details := make([]map[string]interface{}, 1)
+			details[0] = map[string]interface{}{}
```
jharley, line 170: I dislike this but I couldn't wrap my head around a nicer way

brookesargent (approved):
This makes sense to me! And I do like the way notification_details reads in the examples

PR #195, opened 2022-07-25
CI Cleanup

jharley:
A first pass at tidying up the GH Workflows (no more double runs racing with conflicting resource names), testing config setup, and brings some of the Provider boilerplate up to match HashiCorp's "scaffolding" repo.


jharley:
In order to enable provider Matrix tests (a "best practise") two test runs must need to be able run in parallel without generating conflicting resources. This isn't currently the case but I think could be done without a ton of effort.

dstrelau (approved):

PR #197, opened 2022-07-25
CI: fix code coverage

jharley:
Noticed only one of the two files was being processed. Tipped off by https://github.com/codecov/codecov-action/issues/325 that allowing automatic detection fixed their use and it appears to here as well?!


jharley:
Hey, @mitchell-codecov! Thanks for digging into it.. I missed the trailing whitespace but I'm sure I wouldn't have been the last either

dstrelau (approved):

jharley:
Introduced by #195

PR #198, opened 2022-07-27
[columns] introducing `created_at`, `updated_at` and `last_written_at`

mjayaram:
Adds additional fields available when working with Honeycomb Columns


jharley (approved):
🌮✨

Issue #200, opened 2022-07-28
Terraform is unable to apply trigger with the change in recipients with different types


aelmekeev:
**Versions**
0.8.0

**Steps to reproduce**

1. Start with
```hcl
data "honeycombio_query_specification" "example" {
  calculation {
    op = "COUNT"
  }

  filter {
    column = "trace.parent_id"
    op     = "does-not-exist"
  }
}

resource "honeycombio_query" "example" {
  dataset    = local.dataset
  query_json = data.honeycombio_query_specification.example.json
}

data "honeycombio_recipient" "example" {
  type    = "slack"

  detail_filter {
    name  = "channel"
    value = "#alex-test"
  }
}

data "honeycombio_recipient" "example_slack" {
  type    = "slack"

  detail_filter {
    name  = "channel"
    value = "#test_channel"
  }
}

data "honeycombio_recipient" "example_pd" {
  type    = "pagerduty"

  detail_filter {
    name  = "integration_name"
    value = "Alex Test1"
  }
}

resource "honeycombio_trigger" "example" {
  name        = "Alex test"
  description = "Alex test"

  query_id = honeycombio_query.example.id
  dataset  = local.dataset

  frequency = 1800

  threshold {
    op    = ">"
    value = 1000000000
  }

  recipient {
    id = data.honeycombio_recipient.example.id
  }

  recipient {
    id = data.honeycombio_recipient.example_slack.id
  }

  recipient {
    id = data.honeycombio_recipient.example_pd.id

    notification_details {
      pagerduty_severity = "critical"
    }
  }
}
```

```
Apply complete! Resources: 2 added, 0 changed, 0 destroyed.
```

2. Comment out second recipient and apply:

```hcl
  # recipient {
  #   id = data.honeycombio_recipient.example_slack.id
  # }
```

```
Apply complete! Resources: 0 added, 1 changed, 0 destroyed.
```

3. Run apply again:

We would expect to have 0 changes but... 🐛 

```
Apply complete! Resources: 0 added, 1 changed, 0 destroyed.
```

4. Uncomment the previously commented block and run apply again 💣 

```hcl
honeycombio_trigger.example: Modifying... [id=pVVZZxandMW]
╷
│ Error: 400 Bad Request: PagerDuty target not found
│ 
│   with honeycombio_trigger.example,
│   on test.tf line 139, in resource "honeycombio_trigger" "example":
│  139: resource "honeycombio_trigger" "example" {
│ 
```

[...]

PR #202, opened 2022-07-31
bugfix: infinite diff. for burn alert and trigger pagerduty recipients if severity not specified

jharley:
Discovered while investigating #200

### DIFF
--- a/honeycombio/resource_burn_alert_test.go
+++ b/honeycombio/resource_burn_alert_test.go
@@ -135,10 +135,7 @@ resource "honeycombio_burn_alert" "test" {
 
   recipient {
     id = honeycombio_pagerduty_recipient.test.id
-
-    notification_details {
-      pagerduty_severity = "critical"
-    }
+    // default value is critical
   }
 }`, dataset, slo.ID),
 				Check: resource.ComposeTestCheckFunc(
--- a/honeycombio/resource_burn_alert_test.go
+++ b/honeycombio/resource_burn_alert_test.go
@@ -135,7 +135,7 @@ resource "honeycombio_burn_alert" "test" {
 
   recipient {
     id = honeycombio_pagerduty_recipient.test.id
-    // default value is critical
+    // default severity is critical
   }
 }`, dataset, slo.ID),
 				Check: resource.ComposeTestCheckFunc(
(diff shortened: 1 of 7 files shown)


honeycombio/resource_burn_alert_test.go
```diff
@@ -135,10 +135,7 @@ resource "honeycombio_burn_alert" "test" {
 
   recipient {
     id = honeycombio_pagerduty_recipient.test.id
-
-    notification_details {
-      pagerduty_severity = "critical"
-    }
+    // default value is critical
```
dstrelau, line 138: now that you removed the details block, "value" is unclear. Maybe just expand this comment a bit?

honeycombio/type_helpers.go
```diff
@@ -156,10 +156,10 @@ func tpmToFloat(t int) float64 {
 }
 
 func flattenNotificationRecipients(rs []honeycombio.NotificationRecipient) []map[string]interface{} {
-	result := make([]map[string]interface{}, len(rs))
+	result := []map[string]interface{}{}
```
dstrelau, line 159: Why did you change this slice construction? The `make` that was here is more idiomatic, IMO.

honeycombio/type_helpers.go
```diff
@@ -193,11 +194,19 @@ func expandNotificationRecipients(s []interface{}) []honeycombio.NotificationRec
...
-				recipients[i].Details = &honeycombio.NotificationRecipientDetails{
+				rcpt.Details = &honeycombio.NotificationRecipientDetails{
 					PDSeverity: honeycombio.PagerDutySeverity(s.(string)),
 				}
 			}
+		} else {
+			// set default notification details
+			if rcpt.Type == honeycombio.RecipientTypePagerDuty {
+				rcpt.Details = &honeycombio.NotificationRecipientDetails{
+					PDSeverity: honeycombio.PDSeverityCRITICAL,
```
dstrelau, line 205: Where is this used such that we need to hard-code the default here. It'd be nice if we could rely on the API to provide this and not have to duplicate that logic.

honeycombio/type_helpers.go
jharley: Ah, fair. I overcorrected for inserting by index (instead of append). I can flip this back 👍🏻

honeycombio/type_helpers.go
jharley: Nice catch! This is basically unused because we're already reading back from the API when the function is called

dstrelau (approved):

PR #203, opened 2022-08-03
Remove Board inline queries support from client and deprecate board queries' 'dataset'

jharley:


client/query_spec_test.go
```diff
@@ -70,140 +68,6 @@ func TestQuerySpec(t *testing.T) {
 		Granularity: IntPtr(60),   // 1 minute
 	}
 
-	b := &Board{
```
jharley, line 73: In addition to using the Boards API (instead of the Queries API which felt equally find as a smoketest?) these operation specific tests below felt like they were covered in the main test anyway?

dstrelau (approved):

PR #217, opened 2022-08-18
(WIP) add dataset definitions

reulan:
Adding terraform provider equivalents for the  the dataset defintions work that's being added into shepherd.


reulan:
Needs the dataset definitions API to go live, currently its hidden behind a permssions flag not available on prod. once this is added I can make the CI API key have this permission so the 403 permission denied errors can resolve.

jharley:
I can take a look in later today.

> Needs the dataset definitions API to go live, currently its hidden behind a permssions flag not available on prod.

Can we not flag in the Terraform test team?

PR #219, opened 2022-09-01
docs: add clarifying note about Trigger time_range vs frequency

jharley:
This has come up a few times, so I wanted to (attempt to) clarify


josslynzhang (approved):

brookesargent (approved):
Love it! Correlating query duration to time_range is super helpful.

PR #222, opened 2022-09-14
Molly.tf provider experiment

mollystamos123:


client/dataset_definitions.go
```diff
@@ -98,6 +98,8 @@ func (s *datasetDefinitions) Delete(ctx context.Context, dataset string) error {
 	if definition.DurationMs.Name == "" {
 		// valid
 	}
+	var dd DatasetDefinition
+	err := s.client.performRequest(ctx, "PATCH", fmt.Sprintf("/1/dataset_definitions/%s", urlEncodeDataset(dataset)), definition, &dd)
```
reulan, line 102: ahh, good catch for the actual write to honeycomb - I don't think I was ever resetting this value after delete externally

honeycombio/resource_dataset_definition.go
```diff
@@ -111,60 +113,61 @@ func resourceDatasetDefinitionDelete(ctx context.Context, d *schema.ResourceData
 
 // Convert to Terraform Format
 func flattenDatasetDefinition(dd *honeycombio.DatasetDefinition) []map[string]interface{} {
+	fmt.Println("\n\nflattenDatasetDefinition")
 	result := make([]map[string]interface{}, 0)
 
 	// for each field allowed unpack the values and set
-	if dd.DurationMs.Name != "" {
+	if dd.DurationMs.Name != "" && !CheckDatasetDefinitionDurationMs(dd.DurationMs.Name) {
```
reulan, line 120: re-adding the additional matching field validation 👍

PR #229, opened 2022-10-14
bugfix: allow dataset resources to be updated

jharley:
I typoed a dataset description and was reminded that this bug was fixed server side but not in the provider.

### DIFF
--- a/client/dataset_test.go
+++ b/client/dataset_test.go
@@ -12,27 +12,11 @@ func TestDatasets(t *testing.T) {
 
 	c := newTestClient(t)
 	datasetName := testDataset(t)
-	updatedDescription := "buzzing with data"
-	updatedExpandJSONDepth := 3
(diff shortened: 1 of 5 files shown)


client/dataset_test.go
```diff
@@ -12,28 +12,12 @@ func TestDatasets(t *testing.T) {
...
-	updatedDescription := "buzzing with data"
-	updatedExpandJSONDepth := 3
 
 	currentDataset := &Dataset{
 		Name: datasetName,
 		Slug: urlEncodeDataset(datasetName),
 	}
 
-	// create a new dataset with the parameters above
-	t.Run("Create", func(t *testing.T) {
```
jharley, line 24: There were two "Create" tests: this was basically a dupe

brookesargent (approved):
✨

PR #234, opened 2022-10-31
bugfix: missing 'calculation' in Query Specification can cause infinite diff

jharley:
A `honeycombio_query_specification` without a `calculation` will get the default calculation (`COUNT`) returned by the API but cause an infinite diff for resources relying on the JSON input (e.g. `honeycombio_query`).

This resolves this issue, while also marking the `json` attribute of the data source read-only as indicated by the Terraform Plugin SDK documentation.


mjayaram (approved):
🎉

jharley:
Closes #238

PR #242, opened 2022-11-30
column: deprecate 'key_name' in favor of 'name'

jharley:
Using 'key_name' felt like we were leaking an API abstraction into the resource needlessly.

This PR also removes a surprising behavior in the column resource where an existing column would silently be used on create instead of erroring.

### DIFF
--- a/honeycombio/resource_column.go
+++ b/honeycombio/resource_column.go
@@ -23,11 +23,23 @@ func newColumn() *schema.Resource {
 		},
 
 		Schema: map[string]*schema.Schema{
-			"key_name": {
+			"name": {
 				Type:         schema.TypeString,
-				Required:     true,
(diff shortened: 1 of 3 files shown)


honeycombio/resource_column.go
```diff
@@ -91,30 +103,29 @@ func resourceColumnCreate(ctx context.Context, d *schema.ResourceData, meta inte
 	client := meta.(*honeycombio.Client)
 
 	dataset := d.Get("dataset").(string)
-	column := readColumn(d)
-
-	existing, err := client.Columns.GetByKeyName(ctx, dataset, column.KeyName)
```
jharley, line 96: I'm happy to break this out into a separate PR if we feel that's better

dstrelau (approved):

PR #253, opened 2022-12-14
docs: fix README.md markdown rendering

marceloboeira:

The general description includes multiple sections which were rendering in a single line instead of multiple lines because of the missing line in-between

before
Screenshot 2022-12-14 at 16 47 13

after
Screenshot 2022-12-14 at 16 47 01


jharley (approved):
@marceloboeira thanks for this. Agreed that it's more readable this way.

PR #258, opened 2023-01-04
honeycombio_column: delete column on destroy

jharley:
Since the provider's inception this resource treated deletes as a 'noop' -- simply removing the column from state.

In 0.12.0 we removed the silent import of an existing column on create which has forced this gap into the light.

Given that column deletes are fully supported in the API, it felt proper to enable them here as it is assumed that the operator understands that there is removal.

---
  - https://app.asana.com/0/0/1203644387911130

### DIFF
--- a/docs/resources/column.md
+++ b/docs/resources/column.md
@@ -1,8 +1,9 @@
 # Resource: honeycombio_column
 
-Creates a column in a dataset.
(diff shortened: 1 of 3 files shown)


mjayaram (approved):
🎉

PR #272, opened 2023-02-03
correct time_range in honeycombio_trigger example

### LINKED ISSUE #271: honeycombio_trigger example in docs fails at `terraform apply` due to time_range and frequency values
**Versions**
0.13.0

**Steps to reproduce**

1. Try to use the example code in the `honeycombio_trigger` documentation
2. Run `terraform plan` successfully
3. Run `terraform apply`
4. See the error from the Honeycomb API that a query duration can't be 4 times longer than the trigger frequency.
5. See that the default `time_range` on `honeycombio_query_specification` is 7200, which is 12 times longer than the 600 second frequency on the example trigger

**Additional context**

jmhodges-color:

- Closes #271 

When you use the example code in `honeycombio_trigger`'s documentation,
it will fail at time of `terraform apply`.

The error returned from the Honeycomb API says that the query duration
can't be more than 4x greater than the trigger's frequency.

The default time_range on `honeycombio_query_specification` is 7200
which is 12x larger than the `frequency` on `honeycombio_trigger`.

An alternative fix would be to change the `frequency` setting, instead.
I'd be willing to go that route, and produce another PR! It felt like
maybe it's worth mentioning the `time_range` value explicitly here, but
I'm not convinced on that.

Another possible change would be to change the default `time_range`, but
that would be a breaking change. Not only because of user expectations
but also because the provider's current trigger API requires passing
query ids from immutable query objects. If the `time_range` was updated,
it would cause the `honeycombio_query` resources to attempt to update
themselves which would then fail because queries are immutable in the
Honeycomb API.

Trying out the example code in a new terraform configuration

### DIFF
--- a/docs/resources/trigger.md
+++ b/docs/resources/trigger.md
@@ -21,6 +21,8 @@ data "honeycombio_query_specification" "example" {
     column = "trace.parent_id"
     op     = "does-not-exist"
   }
+
+  time_range = 1800
 }
 
 resource "honeycombio_query" "example" {


jharley:
I like this solution: thanks for the thoughtful submission.

FWIW, I am hoping that the [Terraform Plugin Framework](https://github.com/hashicorp/terraform-plugin-framework) will make doing this kind of validation much easier than it is in the current v2 Plugin SDK

jharley (approved):

jmhodges-color:
Oh! Is that migration ongoing or still off in the future?

jharley:
> Oh! Is that migration ongoing or still off in the future?

It's being considered as a possible avenue to resolve some of the open issues. The new framework was written to support co-existing with the current plugin SDK.

PR #291, opened 2023-04-14
Add overlaid charts support

jbalboni:

You can't set a query to display with overlaid aggregates on a board without access to the right graph settings.

Adds the overlaid_charts graph setting that is supported by the API

Configure a board with the overlaid charts option and verify that that is reflected in Honeycomb

### DIFF
--- a/docs/resources/board.md
+++ b/docs/resources/board.md
@@ -96,36 +96,39 @@ resource "honeycombio_board" "overview" {
 
 The following arguments are supported:
 
-* `name` - (Required) Name of the board.
-* `description` - (Optional) Description of the board. Supports markdown.
-* `column_layout` - (Optional) the number of columns to layout on the board, either `multi` (the default) or `single`. Only `visual` style boards (see below) have a column layout.
-* `style` - (Optional) How the board should be displayed in the UI, either `list` (the default) or `visual`.
-* `query` - (Optional) Zero or more configurations blocks (described below) with the queries of the board.
+-   `name` - (Required) Name of the board.
+-   `description` - (Optional) Description of the board. Supports markdown.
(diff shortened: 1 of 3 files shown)


docs/resources/board.md
```diff
@@ -96,36 +96,39 @@ resource "honeycombio_board" "overview" {
 
 The following arguments are supported:
 
-* `name` - (Required) Name of the board.
-* `description` - (Optional) Description of the board. Supports markdown.
-* `column_layout` - (Optional) the number of columns to layout on the board, either `multi` (the default) or `single`. Only `visual` style boards (see below) have a column layout.
-* `style` - (Optional) How the board should be displayed in the UI, either `list` (the default) or `visual`.
-* `query` - (Optional) Zero or more configurations blocks (described below) with the queries of the board.
+-   `name` - (Required) Name of the board.
```
jharley, line 99: This is a bit of a nit, and I'm sure your editor just did this on save, but none of the other files in the repo are formatted this way so if we could keep the `- ` for now that would be ideal

honeycombio/resource_board.go
```diff
@@ -279,6 +285,10 @@ func expandBoardQueryGraphSettings(gs interface{}) (honeycombio.BoardGraphSettin
 		graphSettings.UseUTCXAxis = true
 	}
```
jharley, line 287:
```suggestion
```

jharley (changes requested):
Thanks for this, @jbalboni!

Looks great with the exception of the Markdown formatting I mention below (in the name of consistency).

Two other suggestions as well but both (very) minor.

jharley (approved):
Thanks! Looks great 🚀

PR #297, opened 2023-04-20
Add honeycombio_column and columns data sources

### LINKED ISSUE #42: Data source: column
_It's technically possible to create a data source to query columns of a dataset (#41), but I'm not sure if this is useful. If you could use this, please leave a comment! 🙂_

irvingpop:

- Closes #42 

This PR adds a minimal implementation for the `honeycombio_column` and `honeycombio_columns` data sources

after running `make install_macos`,  try out the following TF for fun:

```hcl
data "honeycombio_columns" "all" {
  dataset = var.required_columns_dataset_name
}

# Iterate through the list of retrieved columns
data "honeycombio_column" "all" {
  for_each = toset(data.honeycombio_columns.all.names)

  dataset = var.required_columns_dataset_name
  name    = each.key
}

# Retrieve the details of a single column
data "honeycombio_column" "mycol" {
  dataset = var.required_columns_dataset_name
  name    = "db.type"
}

# Ensure a column in another dataset remains in sync
resource "honeycombio_column" "othercol" {
  dataset     = var.required_columns_dataset_name
  name        = "othercol"
  type        = data.honeycombio_column.mycol.type
  description = data.honeycombio_column.mycol.description
  hidden      = data.honeycombio_column.mycol.hidden

  # but only if its updated_at is greater than an arbitrary date
  count = timecmp(data.honeycombio_column.mycol.updated_at, "2023-04-19T00:00:00Z") == 1 ? 1 : 0
}

output "allcols" {
  value = data.honeycombio_columns.all.names
}

output "mycol" {
  value = data.honeycombio_column.all["db.type"]
}
```

### DIFF
--- a/docs/data-sources/column.md
+++ b/docs/data-sources/column.md
@@ -1,0 +1,49 @@
+# Data Source: honeycombio_column
+
+The column data source retrieves the details of a single column
+
+## Example Usage
+
+```hcl
+variable "dataset" {
+  type = string
+}
+
+# Retrieve the details of a single column
+data "honeycombio_column" "mycol" {
+  dataset = var.dataset
+  name    = "mycol"
+}
+
+# Ensure a column in another dataset remains in sync
+resource "honeycombio_column" "othercol" {
+  dataset     = "otherdataset"
+  name        = data.honeycombio_column.mycol.name
+  type        = data.honeycombio_column.mycol.type
+  description = data.honeycombio_column.mycol.description
+  hidden      = data.honeycombio_column.mycol.hidden
+
+  # but only if its updated_at is greater than an arbitrary date
+  count = timecmp(data.honeycombio_column.mycol.updated_at, "2023-04-20T00:00:00Z") == 1 ? 1 : 0
+}
+
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `dataset` - (Required) The dataset this column is associated with
+* `name` - (Required) The name of the column
+
+## Attribute Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `name` - the name of the column
+* `type` - the type of the column (string, integer, float, or boolean)
+* `description` - the description of the column
+* `hidden` - whether or not the column is hidden from the query builder and results
+* `last_written` - the timestamp (string) the last time the column received data
+* `created_at` - the timestamp (string) when the column was created
+* `updated_at` - the timestamp (string) when the column's metadata (type, description, etc) was last changed
--- a/docs/data-sources/column.md
+++ b/docs/data-sources/column.md
@@ -1,6 +1,9 @@
 # Data Source: honeycombio_column
 
-The column data source retrieves the details of a single column
+The `honeycombio_column` data source retrieves the details of a single column in a dataset.
+
+-> **Note** Terraform will fail unless a column is returned by the search. Ensure that your search is specific enough to return a column.
+If you want to match multiple columns, use the `honeycombio_columns` data source instead.
 
 ## Example Usage
 
@@ -14,19 +17,6 @@ data "honeycombio_column" "mycol" {
   dataset = var.dataset
   name    = "mycol"
 }
-
-# Ensure a column in another dataset remains in sync
-resource "honeycombio_column" "othercol" {
-  dataset     = "otherdataset"
-  name        = data.honeycombio_column.mycol.name
-  type        = data.honeycombio_column.mycol.type
-  description = data.honeycombio_column.mycol.description
-  hidden      = data.honeycombio_column.mycol.hidden
-
-  # but only if its updated_at is greater than an arbitrary date
-  count = timecmp(data.honeycombio_column.mycol.updated_at, "2023-04-20T00:00:00Z") == 1 ? 1 : 0
-}
-
 ```
 
 ## Argument Reference
@@ -40,10 +30,9 @@ The following arguments are supported:
 
 In addition to all arguments above, the following attributes are exported:
 
-* `name` - the name of the column
 * `type` - the type of the column (string, integer, float, or boolean)
 * `description` - the description of the column
 * `hidden` - whether or not the column is hidden from the query builder and results
-* `last_written` - the timestamp (string) the last time the column received data
-* `created_at` - the timestamp (string) when the column was created
-* `updated_at` - the timestamp (string) when the column's metadata (type, description, etc) was last changed
+* `last_written_at` - the ISO8601 formatted time that the column last received data
+* `created_at` - the ISO8601 formatted time when the column was created
+* `updated_at` - the  ISO8601 formatted time when the column's metadata (type, description, etc) was last changed
--- a/docs/data-sources/columns.md
+++ b/docs/data-sources/columns.md
@@ -1,0 +1,73 @@
+# Data Source: honeycombio_columns
+
+The columns data source allows the columns of a dataset to be retrieved.
+
+## Example Usage
+
+```hcl
+variable "dataset" {
+  type = string
+}
+
+# returns all columns
+data "honeycombio_columns" "all" {
+  dataset = var.dataset
+}
+
+# only returns the columns starting with 'foo_'
+data "honeycombio_columns" "foo" {
+  dataset     = var.dataset
+  starts_with = "foo_"
+}
+
+# Iterate through the list of retrieved columns
+data "honeycombio_column" "all" {
+  for_each = toset(data.honeycombio_columns.all.names)
+
+  dataset = var.dataset
+  name    = each.key
+}
+
+# And you can now use this to read data.honeycombio_column.all["mycolumn"].type
+
+####################################################
+# Example: Create only the missing columns
+####################################################
+locals {
+  required_columns = {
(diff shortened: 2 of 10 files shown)


.github/workflows/ci.yaml
```diff
@@ -55,6 +55,7 @@ jobs:
         env:
           HONEYCOMB_API_KEY: ${{ secrets.HONEYCOMB_API_KEY }}
           HONEYCOMB_DATASET: testacc
+          HONEYCOMB_COLUMN: duration_ms
```
jharley, line 58: I'd rather we _not_ rely on the environment here: it looks like you're just mirroring the pattern put in place for the `honeycombio_datasets` datasource but datasets are a bit of a special case because we don't (generally) allow deletion of datasets to allow tests to create and cleanup after themselves.

I'd rather see the test for this datasource make a few columns at the start of the test and remove them at the end of the test. The test for the `honeycombio_recipients` datasource has a [decent example of this](https://github.com/honeycombio/terraform-provider-honeycombio/blob/main/honeycombio/data_source_recipients_test.go#L43-L56).

honeycombio/data_source_columns.go
```diff
@@ -0,0 +1,90 @@
...
+				Optional: true,
+			},
+			"names": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+			"types": {
```
jharley, line 35: I don't love the idea of a separate list of types where we hope that the two lists are sharing the same index.

What I'd like to see instead is the pattern emerging in the community where datasources that are plural return a list of identifiers (names in this case) which can then be passed to a singular data source to get a fully enriched object with all of the properties set. In this case, that would mean the addition of a `honeycombio_column` data source which takes a name.

This would allow something like this to give full objects with all the fields set:

```hcl
data "honeycombio_columns" "all" {
  dataset = var.dataset
}

data "honeycombio_column" "all" {
  for_each = data.honeycombio_columns.names
  
  dataset = var.dataset
}
```

And you can now use this to check `data.honeycombio_column.all["mycolumn"].type`.

This does mean a bit more code (with the second datasource addition), but I think the flexibility is worth it and it plays nicely with the emerging idiom in the community.

honeycombio/data_source_columns.go
```diff
@@ -0,0 +1,90 @@
...
+		switch *d.Type {
+		case honeycombio.ColumnTypeString:
+			types[i] = "string"
+		case honeycombio.ColumnTypeFloat:
+			types[i] = "float"
+		case honeycombio.ColumnTypeInteger:
+			types[i] = "integer"
+		case honeycombio.ColumnTypeBoolean:
+			types[i] = "boolean"
+		}
```
jharley, line 82: I think we should drop the `types` list as mentioned above, but you can avoid the switch here altogether as the underlying type is a string anyway :)

```suggestion
        types[i] = *d.Type.(string)
```

docs/data-sources/columns.md
```diff
@@ -0,0 +1,29 @@
...
+## Example Usage
+
+```hcl
+# returns all columns
+data "honeycombio_columns" "all" {
+}
+
+# only returns the columns starting with 'foo_'
+data "honeycombio_columns" "foo" {
+  starts_with = "foo_"
```
jharley, line 14:
```suggestion
  dataset     = var.dataset
  starts_with = "foo_"
```

docs/data-sources/columns.md
```diff
@@ -0,0 +1,29 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `starts_with` - (Optional) Only return columns starting with the given value.
```
jharley, line 22:
```suggestion
* `dataset` - (Required) The dataset this query is associated with.
* `starts_with` - (Optional) Only return columns starting with the given value.
```

jharley (changes requested):
Thanks for this, @irvingpop!

Left some feedback below about the direction I'd like to see this take. Also, happy to hop on a quick call to work through any local dev issues

.github/workflows/ci.yaml
irvingpop: Great points and example! fixing.

honeycombio/data_source_columns.go
irvingpop: okay phew, I was really hating that output 🙂 
will implement a `honeycombio_column` datasource as well

honeycombio/data_source_column.go
```diff
@@ -0,0 +1,102 @@
...
+
+func dataSourceHoneycombioColumnRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*honeycombio.Client)
+
+	dataset := d.Get("dataset").(string)
+	matchName := d.Get("name").(string)
+
+	column, err := client.Columns.GetByKeyName(ctx, dataset, matchName)
+	if err != nil {
+		return diag.FromErr(fmt.Errorf("column not found, API returned: %s", err.Error()))
```
irvingpop, line 89: I'm wondering if this is the idiomatic behavior here or not?   Currently this makes TF bomb out if you try to get data on a column which doesn't exist - is that right, or would it be better to return _something_?

jharley: There are both approaches out there in the ecosystem.

So far in this provider the plural version of a datasource will happily return an empty list (no failure), but the singular fails if you don't return a match.

I think this makes sense given that you're asking for a specific thing with the singular and if you don't get it the only behaviour that seems to make sense is to raise an error (e.g. "I failed to lookup what you asked me for").

That said, I'm open to alternatives :) What's the use case for returning nothing here and potentially having weird errors when trying to access the datasource?

irvingpop:
heh, in testing I managed to do a horrible thing:
```
allcols = tolist([
  " db.type",
  " duration_ms",
  " error",
  "db.system",
  "db.type",
  "duration_ms",
  "error",
  "http.flavor",
```

if you try to access one of those columns, like
```
data "honeycombio_column" "first" {
  dataset = var.required_columns_dataset_name
  name = data.honeycombio_columns.all.names[0]
}
```

the API throws a 400.   I'm wondering if it's okay to let this exist, or to urlencode the keyname [here](https://github.com/honeycombio/terraform-provider-honeycombio/blob/36432000413c8ac9ab1a43da82fc77d1e04c8747/client/column.go#L97-L98)?

honeycombio/data_source_column.go
irvingpop: You're right,  after thinking it through some more I'm with you

irvingpop:
wow that was fun!  I've put this through the paces locally and I'm feeling pretty good about it. I even have a starter-pack PR ready to go for when this lands 🙂 

I just squash-rebased and it's ready for a re-review.

docs/data-sources/column.md
```diff
@@ -0,0 +1,49 @@
...
+  dataset     = "otherdataset"
+  name        = data.honeycombio_column.mycol.name
+  type        = data.honeycombio_column.mycol.type
+  description = data.honeycombio_column.mycol.description
+  hidden      = data.honeycombio_column.mycol.hidden
+
+  # but only if its updated_at is greater than an arbitrary date
+  count = timecmp(data.honeycombio_column.mycol.updated_at, "2023-04-20T00:00:00Z") == 1 ? 1 : 0
+}
+
```
jharley, line 29: I'd rather see complex/neat examples like this in the [example directory](https://github.com/honeycombio/terraform-provider-honeycombio/tree/main/example) and have the docs focus on day to day usage

```suggestion
```

docs/data-sources/columns.md
```diff
@@ -0,0 +1,72 @@
...
+  for_each = toset(local.cols_to_create)
+
+  name    = each.key
+  type    = lookup(local.required_columns, each.key, "string")
+  dataset = var.dataset
+}
+
+output "columns_created" {
+  value = local.cols_to_create
+}
```
jharley, line 58: As above, I'd rather see this moved into the [example](https://github.com/honeycombio/terraform-provider-honeycombio/tree/main/example) directory

```suggestion
```

jharley (changes requested):
This is looking great and just about ready to go! 🎉 

A few edits requested before merging (largely to match the way existing things are in the repo)

docs/data-sources/column.md
```diff
@@ -0,0 +1,49 @@
...
+The following arguments are supported:
+
+* `dataset` - (Required) The dataset this column is associated with
+* `name` - (Required) The name of the column
+
+## Attribute Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `name` - the name of the column
```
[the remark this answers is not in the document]
irvingpop: Good catch. Does it make sense to leave the lines above it?  (39-41)

irvingpop:
Thanks for the detailed suggestions @jharley !! I just added the examples (moved out from docs) - although now I'm wondering if it's inappropriate to link to the examples folder from the docs to boost visibility of them.

If not and we're good to merge from here, I'll run another squash rebase and push.

jharley (approved):

PR #301, opened 2023-04-24
client - escape query string when listing burn alerts for an SLO

### LINKED ISSUE #300: client: should escape any query params
As reminded by #297, we should always `url.QueryEscape` any query params

jharley:

- Closes #300 

As reported in #297, blindly passing HCL inputs into the query can cause bad behaviour. This is the only other place in the codebase (after #297 is merged) which uses an unescaped query parameter.

---
  - https://app.asana.com/0/0/1204461304887421

### DIFF
--- a/client/burn_alert.go
+++ b/client/burn_alert.go
@@ -3,6 +3,7 @@ package client
 import (
 	"context"
 	"fmt"
+	"net/url"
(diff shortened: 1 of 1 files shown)


brookesargent (approved):
✨

PR #302, opened 2023-04-24
Mention the example directory from the main docs page

irvingpop:
The examples in the [example directory](/example/) are so good, and I didn't even know about them!   Let's make it a tiny bit more visible

Go to [the rendered markdown for the docs index page](https://github.com/honeycombio/terraform-provider-honeycombio/blob/a4b6ab2c112fca063e7a3bae99ed39407b0b661d/docs/index.md) and hit the example link.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -36,6 +36,8 @@ resource "honeycombio_marker" "hello" {
 }
 ```
 
+More advanced examples may be found in the [example directory](/example/).
+
 ## Authentication
 
 The Honeycomb provider requires an API key to communicate with the Honeycomb API. API keys and their permissions can be managed in _Team settings_.


docs/index.md
```diff
@@ -36,6 +36,8 @@ resource "honeycombio_marker" "hello" {
 }
 ```
 
+More advanced examples may be found in the [example directory](/example/).
```
jharley, line 39: As this is most commonly viewed on the [Terraform Registry listing](https://registry.terraform.io/providers/honeycombio/honeycombio/latest/docs), I think this will need to be a direct link
```suggestion
More advanced examples can be found in the [example directory](https://github.com/honeycombio/terraform-provider-honeycombio/tree/main/example).
```

irvingpop: A fair point, that was my first iteration and then I tried to be clever 🙂

jharley (approved):

PR #306, opened 2023-05-11
migrate honeycombio_trigger resource 'as is' to Plugin Framework

jharley:
First pass at migrating the resource: with this (verbose) groundwork laid the test suite can be fleshed out and the open bugs squashed 🤞🏻 

The current SDKv2 based Trigger resource is being left in place (and its tests skipped) just not wired up for use.

### DIFF
--- a/honeycombio/resource_trigger_test.go
+++ b/honeycombio/resource_trigger_test.go
@@ -11,6 +11,7 @@ import (
 )
 
 func TestAccHoneycombioTrigger_basic(t *testing.T) {
+	t.Skip("legacy test, skipping")
 	var triggerBefore, triggerAfter honeycombio.Trigger
 
 	dataset := testAccDataset()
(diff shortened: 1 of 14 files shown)


internal/helper/validation/divisible_by.go
jharley, line 1: The Framework didn't get this validator included which is present in SDKv2.

jharley, line 1: SDKv2 included such a validator, and we used it, but the Plugin Framework does not. This is a reimplementation.

honeycombio/resource_trigger_test.go
jharley, line 1: With the Trigger resource removed from the SDKv2 provider, these tests won't run (they can't find the resource). Leaving them here until all the code is removed, so just skipping for easy-mode.

brookesargent (approved):
🏆

internal/provider/trigger_resource.go
```diff
@@ -0,0 +1,436 @@
...
+func (r *triggerResource) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_trigger"
+}
+
+func (r *triggerResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Manages a Trigger.",
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				Description: "Numeric identifier of the Trigger.",
```
dstrelau, line 53: Numeric?

jharley: 🙃 I think that was boilerplate copy-pasta. Thanks!

PR #313, opened 2023-06-01
honeycombio_trigger: remove SDK-based trigger resource

jharley:
The SDK-based `honeycombio_trigger` resource is no longer used


brookesargent (approved):
best type of PR 🎉

PR #314, opened 2023-06-01
honeycombio_trigger - add evaluation schedule support

jharley:
Adds support for Trigger Evaluation Schedules to the `honeycombio_trigger` resource.

### DIFF
--- a/client/trigger.go
+++ b/client/trigger.go
@@ -59,9 +59,16 @@ type Trigger struct {
 	Query   *QuerySpec `json:"query,omitempty"`
 	QueryID string     `json:"query_id,omitempty"`
 	// Alert Type. Describes scheduling behavior for triggers.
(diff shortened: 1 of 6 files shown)


brookesargent (approved):
🚀

client/trigger.go
```diff
@@ -97,37 +128,6 @@ func TriggerThresholdOps() []TriggerThresholdOp {
 	}
 }
 
-// Allowed values for alert_type. | on_change is default
-const (
-	TriggerAlertTypeValueOnChange string = "on_change"
-	TriggerAlertTypeValueOnTrue   string = "on_true"
-)
-
-func (t *Trigger) MarshalJSON() ([]byte, error) {
```
jharley, line 106: The 'less magic, more native Go' workflow of the Framework (vs the SDK) means this hack is no longer needed ✨

PR #324, opened 2023-06-20
client: update column test to not rename column

jharley:
Functionality about to be removed server side.


client/column_test.go
```diff
@@ -74,6 +73,9 @@ func TestColumns(t *testing.T) {
 		assert.NoError(t, err)
 
 		data.ID = column.ID
+		assert.Equal(t, column.Description, data.Description)
```
jharley, line 76: These arguably should have been asserted previously

mjayaram (approved):

PR #331, opened 2023-07-05
Fix missing recipient panic

LukeCarrier:
It looks like these missing returns are an oversight from #311?

## Which problem is this PR solving?

No bug, but two panics we've observed:

```text
panic: runtime error: index out of range [-1]

goroutine 123 [running]:
github.com/honeycombio/terraform-provider-honeycombio/internal/provider.(*triggerResource).Update(0xc000120f30, {0xeb5980, 0xc0001f6630}, {{{{0xeba508, 0xc000726e10}, {0xc679a0, 0xc000329a70}}, {0xebbd28, 0xc0002da7d0}}, {{{0xeba508, ...}, ...}, ...}, ...}, ...)
	github.com/honeycombio/terraform-provider-honeycombio/internal/provider/trigger_resource.go:477 +0xe1b
github.com/hashicorp/terraform-plugin-framework/internal/fwserver.(*Server).UpdateResource(0xc000161b80, {0xeb5980, 0xc0001f6630}, 0xc000543188, 0xc000775098)
	github.com/hashicorp/terraform-plugin-framework@v1.2.0/internal/fwserver/server_updateresource.go:118 +0x731
github.com/hashicorp/terraform-plugin-framework/internal/fwserver.(*Server).ApplyResourceChange(0xeb5980?, {0xeb5980, 0xc0001f6630}, 0xc000537720, 0xc0007752b8)
	github.com/hashicorp/terraform-plugin-framework@v1.2.0/internal/fwserver/server_applyresourcechange.go:99 +0x1a5
github.com/hashicorp/terraform-plugin-framework/internal/proto5server.(*Server).ApplyResourceChange(0xc000161b80, {0xeb5980?, 0xc0001f64e0?}, 0xc0005376d0)
	github.com/hashicorp/terraform-plugin-framework@v1.2.0/internal/proto5server/server_applyresourcechange.go:52 +0x27b
github.com/hashicorp/terraform-plugin-mux/tf5muxserver.muxServer.ApplyResourceChange({0xc000239f50, 0xc000239fb0, {0xc000133c20, 0x2, 0x2}, {0x0, 0x0, 0x0}, {0x0, 0x0, ...}, ...}, ...)
	github.com/hashicorp/terraform-plugin-mux@v0.10.0/tf5muxserver/mux_server_ApplyResourceChange.go:27 +0x102

Error: The terraform-provider-honeycombio_v0.15.1 plugin crashed!
```

```text
panic: runtime error: index out of range [-1]

goroutine 1257 [running]:
[...]

PR #339, opened 2023-08-05
bugfix - resolve trigger recipient panics by relying on config vs plan

### LINKED ISSUE #309: PagerDuty notification recipients don't handle default computed value
**Versions**

0.15.0 or newer

**Steps to reproduce**

Effects `honeycombio_trigger` resources.

If a configured recipient (by ID or Type) is a PagerDuty recipient and doesn't provide a severity in the optional `notification_details` block, 

```hcl
data "honeycombio_query_specification" "test" {
  calculation {
    op     = "AVG"
    column = "duration_ms"
  }
  time_range = 1200
}

resource "honeycombio_query" "test" {
  dataset    = var.dataset
  query_json = data.honeycombio_query_specification.test.json
}

resource "honeycombio_pagerduty_recipient" "test" {
  integration_key  = var.pagerduty_integration_key
  integration_name = "severity-repro"
}

resource "honeycombio_trigger" "test" {
  name    = "Test trigger from terraform-provider-honeycombio"
  dataset = var.dataset

  description = "My nice description"

  query_id = honeycombio_query.test.id

  threshold {
    op    = ">"
    value = 100
  }

  frequency = data.honeycombio_query_specification.test.time_range / 2

  recipient {
    id = honeycombio_pagerduty_recipient.test.id
  }
}
```

Running the code above produces the following error on `apply`:

```
Error: Provider produced inconsistent result after apply
        
When applying changes to honeycombio_trigger.test, provider
"provider[\"registry.terraform.io/hashicorp/honeycombio\"]" produced an
unexpected new value: .recipient: planned set element
cty.ObjectVal(map[string]cty.Value{"id":cty.StringVal("FTh7K9xqTgS"),
"notification_details":cty.ListValEmpty(cty.Object(map[string]cty.Type{"pagerduty_severity":cty.String})),
"target":cty.UnknownVal(cty.String), "type":cty.UnknownVal(cty.String)}) does
not correlate with any element in actual.
```

**Workaround**

Until this is resolved, all PagerDuty `recipient` blocks should have `notification_details` provided -- even if you're relying on the default severity level of `critical`.

```hcl
  recipient {
    id = honeycombio_pagerduty_recipient.test.id

    notification_details {
      pagerduty_severity = "critical"
    }
  }
```

**Additional context**

Introduced by #306

Updating the custom [plan modifier](https://developer.hashicorp.com/terraform/plugin/framework/migrating/resources/plan-modification) for the `recipient` block should solve resolve this "computed default sometimes" case.

jharley:
Fixes the currently known pair of panics on the `honeycombio_trigger` resource by relying on config vs plan as the source of what to persist to state.

- Resolves #309
- Replaces #331

### DIFF
--- a/internal/provider/trigger_resource_test.go
+++ b/internal/provider/trigger_resource_test.go
@@ -69,6 +69,7 @@ func TestAcc_TriggerResourceUpgradeFromVersion014(t *testing.T) {
 			{
 				ProtoV5ProviderFactories: testAccProtoV5MuxServerFactory,
 				Config:                   config,
+				PlanOnly:                 true,
 			},
 		},
 	})
@@ -352,8 +353,6 @@ resource "honeycombio_trigger" "test" {
 }
 
 func TestAcc_TriggerResourcePagerDutyUnsetSeverity(t *testing.T) {
-	t.Skip("known issue: see https://github.com/honeycombio/terraform-provider-honeycombio/issues/309")
-
 	dataset := testAccDataset()
 
 	resource.Test(t, resource.TestCase{
@@ -390,7 +389,7 @@ resource "honeycombio_trigger" "test" {
 
   threshold {
     op    = ">"
-    value = 100
+    value = 1 - 0.99
   }
 
   frequency = data.honeycombio_query_specification.test.time_range / 2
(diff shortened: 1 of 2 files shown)


internal/provider/trigger_resource_test.go
```diff
@@ -405,49 +404,6 @@ resource "honeycombio_trigger" "test" {
 	})
 }
 
-func TestAcc_TriggerResourceThresholdFloatingPointBug(t *testing.T) {
```
jharley, line 408: Rather than keeping this as a dedicated test I just rolled the important bit (the floating point math) into the test above it.

internal/provider/trigger_resource_test.go
```diff
@@ -69,6 +69,7 @@ func TestAcc_TriggerResourceUpgradeFromVersion014(t *testing.T) {
 			{
 				ProtoV5ProviderFactories: testAccProtoV5MuxServerFactory,
 				Config:                   config,
+				PlanOnly:                 true,
```
jharley, line 72: This was always the recommendation of how this test should be shaped but the way things were working broke it 🙃 

Restored to the documented approach

dstrelau (approved):

PR #372, opened 2023-10-10
client: improve error handling and reporting

### LINKED ISSUE #245: query_result datasource should retry-and-backoff on 429 response
As the Query Data API has rate limits we should handle these retries transparently in the client up to a certain timeout.

jharley:
An overhaul of the provider's embedded API client with the goal of making it more resilient to transient errors (429, 502, 504) and doing a better job of making error details available when an unrecoverable error happens.

The retries brought to us by the standard (in Terraform land) [go-retryablehttp](https://github.com/hashicorp/go-retryablehttp) in a fairly simplistic initial setup[1].

Richer error messages brought in by trying to make as much use as possible of RFC7807-style errors coming back from our API and then wiring them into detailed errors (`Diagnostics`) during Terraform runs. 

- Closes #245

1: very open to feedback on the max retries, the min and max values ([link](https://github.com/honeycombio/terraform-provider-honeycombio/pull/372/files#diff-bd3a55a72186f59e2e63efb4951573b2f9e4a7cc98086e922b0859f8ccc1dd09R120-R122)). They feel safe and conservative to me


internal/helper/diag.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// AddDiagnosticOnError is a helper function which will take an error
+// and a context-specific summary of the action.
+//
+// If err is nil, this function will return false.
+// If err is non-nil, this function will return ture and take one the following actions:
+//   - if it is a client.DetailedError, a DetailedErrorDiagnostic will be added to diag.
+//   - otherwise a generic error diagnostic will be added to diag.
+func AddDiagnosticOnError(diag *diag.Diagnostics, summary string, err error) bool {
```
jharley, line 38: I attempted to come up with a signature for this helper that could expose the 404 case needed to be handled on all the `Read` methods for resources and didn't love anything I came up with.

cewkrupa: the best thing I can think of would be passing in a callback func to the function, like

```golang
trigger, err := r.client.Triggers.Get(ctx, state.Dataset.ValueString(), state.ID.ValueString())
if helper.AddDiagnosticOnError(&resp.Diagnostics, "Creating Honeycomb Trigger", err, func(detailedErr *client.DetailedError) bool {
	if detailedErr.IsNotFound() {
                // not super sure if this works with Go's closures like it would if this was JS
		resp.State.RemoveResource(ctx)
		return false
	}
	return true
}) {
  return
}

// ...

func AddDiagnosticOnError(diag *diag.Diagnostics, summary string, err error, callback func(detailedErr *client.DetailedError) bool) bool {
	if err == nil {
		return false
	}

	var detailedErr *client.DetailedError
	if errors.As(err, &detailedErr) {
		// the return val of callback determines if we try to append the error, letting the caller decide
		if callback(detailedErr) {			
			diag.Append(DetailedErrorDiagnostic{
				summary: "Error " + summary,
				e:       detailedErr,
			})
		}
	} else {
		diag.AddError("Error "+summary, err.Error())
	}
	return true
}
```

...but I don't think it makes it much clearer at all. what you have seems like a great improvement already!

internal/helper/type_strings.go
```diff
@@ -1,25 +1,11 @@
...
-
-	return out
-}
-
-func RecipientTypeStrings(recipientTypes []client.RecipientType) []string {
-	in := recipientTypes
-	out := make([]string, len(in))
-
+// AsStringSlice converts a slice of string-like things to a slice of strings.
+func AsStringSlice[S ~string](in []S) []string {
```
cewkrupa, line 4: yay generics!

client/client.go
```diff
@@ -78,10 +64,39 @@ type Client struct {
...
+		APIKey:     os.Getenv("HONEYCOMB_API_KEY"),
+		APIUrl:     DefaultAPIHost,
+		Debug:      false,
+		HTTPClient: cleanhttp.DefaultPooledClient(),
+		UserAgent:  defaultUserAgent,
+	}
+
+	// if API Key is still unset, try using the legacy environment variable
+	if c.APIKey == "" {
+		c.APIKey = os.Getenv("HONEYCOMBIO_API_KEY")
```
cewkrupa, line 78: nit: what would you think of moving these `HONEYCOMB_API_KEY` & `HONEYCOMBIO_API_KEY` strings into some constants? I definitely thought they were the same value for a solid 5 minutes until I noticed the `IO` in the legacy env var 😅

client/client.go
```diff
@@ -91,17 +106,34 @@ func NewClient(config *Config) (*Client, error) {
...
+	}
+
+	client.httpClient = &retryablehttp.Client{
+		Backoff:      retryablehttp.DefaultBackoff,
+		CheckRetry:   client.retryHTTPCheck,
+		ErrorHandler: retryablehttp.PassthroughErrorHandler,
+		HTTPClient:   cfg.HTTPClient,
+		RetryWaitMin: 200 * time.Millisecond,
+		RetryWaitMax: 10 * time.Second,
+		RetryMax:     15,
```
cewkrupa, line 122: Would we want to make these values configurable somehow?

cewkrupa (approved):
this looks great to me!

client/client.go
jharley: I don't _think_ so at this point but I thought about it as well. If this were a generic client then I think yes, but I think/hope these should be "good enough" (for now, anyway)

dstrelau (approved):

client/client.go
jharley: Done!

PR #386, opened 2023-11-07
Add support for 'msteams' recipients

jharley:
Adds supports for `msteams` recipients.

Also fixes a "sneaky bug" where a previous refactor would allow `marker` type recipients on Burn Alerts during the validation stage (the API still would reject them).

---
  - https://app.asana.com/0/0/1205887299919797

### DIFF
--- a/internal/provider/notification_recipients.go
+++ b/internal/provider/notification_recipients.go
@@ -18,9 +18,9 @@ import (
 	"github.com/honeycombio/terraform-provider-honeycombio/internal/models"
 )
 
-func notificationRecipientSchema() schema.SetNestedBlock {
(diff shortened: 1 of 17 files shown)


internal/provider/notification_recipients.go
```diff
@@ -18,9 +18,9 @@ import (
 	"github.com/honeycombio/terraform-provider-honeycombio/internal/models"
 )
 
-func notificationRecipientSchema() schema.SetNestedBlock {
+func notificationRecipientSchema(allowedTypes []client.RecipientType) schema.SetNestedBlock {
```
jharley, line 21: This fixes the bug introduced by the refactor

cewkrupa: Good catch!!

cewkrupa (approved):
This looks great, thank you!!

PR #392, opened 2023-11-13
client: Fix bug where API errors weren't including the field or getting separated

lafentres:
This PR fixes a bug where the errors returned from the client weren't showing the impacted field and weren't getting separated when there were multiple errors associated with 1 request. This made any server-side errors a little difficult to read/fix.

I'm not super familiar with all the different forms the Details section of DetailedError can take, so if there are checks I can remove, please let me know. 

Pre-requisites:
- You'll need Go 1.20 and the latest Terraform installed.
- You'll need a team in Honeycomb to test against.
- You'll need a dataset, derived column, and SLO for that team.
- You'll need an API key for that team.

#### Setup:
1. Create a new directory called `providers` somewhere you can easily access (but not inside the same folder as your Terraform config).
1. Build the provider by checking out this branch and running `make build`
1. Move the executable created to the `providers` directory you created earlier.
   ```
   mv terraform-provider-honeycombio /PATH/TO/YOUR/DIRECTORY/providers
   ```
1. Add the following to your ~/.terraformrc file:
   ```
   provider_installation {

     # Use /PATH/TO/YOUR/DIRECTORY/providers/terraform-provider-honeycombio
     # as an overridden package directory for the honeycombio/honeycombio provider. 
     # This disables the version and checksum verifications for this provider and 
     # forces Terraform to look for the null provider plugin in the given directory.
     dev_overrides {
       "honeycombio/honeycombio" = "/PATH/TO/YOUR/DIRECTORY/providers"
     }

     # For all other providers, install them directly from their origin provider
     # registries as normal. If you omit this, Terraform will _only_ use
     # the dev_overrides block, and so no other providers will be available.
     direct {}
   }
   ```
1. Create another new directory (all Terraform commands will be run in this directory)
1.  [Download and save this file](https://github.com/honeycombio/terraform-provider-honeycombio/files/13342818/error-display.tf.txt) as `main.tf` in your new directory. You'll be modifying it in the steps below. 
1. Set the following environment variables:
   ```
   export HONEYCOMB_API_KEY=YOUR_API_KEY_HERE
   export TF_VAR_dataset=YOUR_DATASET_SLUG_HERE
   export TF_VAR_slo_id=YOUR_SLO_ID_HERE
   ```
1. Run `terraform init`. You should see a yellow warning block telling you that development overrides are in place. It should look like this:
   Screenshot 2023-11-13 at 3 34 46 PM

#### Provision resources for r/burn_alert:
1. Run `terraform apply`. This should fail with an error that look like the following:
   Screenshot 2023-11-13 at 5 02 16 PM
   - If you check in the UI, you should see that the burn alert was not created.
 
## Related links
- [API documentation](https://docs.honeycomb.io/api/tag/Errors#schema/ValidationError)

### Before:
Screenshot 2023-11-13 at 5 04 57 PM

### After:
Screenshot 2023-11-13 at 5 02 16 PM

### DIFF
--- a/client/errors.go
+++ b/client/errors.go
@@ -18,23 +18,54 @@ type DetailedError struct {
 	// Title is a human-readable summary that explains the type of the problem.
 	Title string `json:"title,omitempty"`
 	// Details is an array of structured objects that give details about the error.
-	Details []struct {
-		Code        string `json:"code"`
-		Description string `json:"description"`
-		Field       string `json:"field"`
-	} `json:"type_detail,omitempty"`
+	Details []ErrorTypeDetail `json:"type_detail,omitempty"`
+}
+
+type ErrorTypeDetail struct {
+	Code        string `json:"code"`
+	Description string `json:"description"`
+	Field       string `json:"field"`
+}
+
+func (td ErrorTypeDetail) String() string {
+	response := ""
+	if td.Code != "" {
+		response += td.Code
+		// If any of the fields that could come next are present, add a space separator
+		if td.Field != "" || td.Description != "" {
+			response += " "
+		}
+	}
+
+	if td.Field != "" {
+		response += td.Field
+		// If any of the fields that could come next are present, add a dash separator
+		if td.Description != "" {
+			response += " - "
+		}
+	}
+
+	if td.Description != "" {
+		response += td.Description
+	}
+
+	return response
 }
 
 // Error returns a pretty-printed representation of the error
 func (e DetailedError) Error() string {
 	if len(e.Details) > 0 {
 		var response string
-		for i, d := range e.Details {
-			response += d.Code + " - " + d.Description
-			if i > len(e.Details)-1 {
-				response += ", "
+
+		for index, details := range e.Details {
+			response += details.String()
+
+			// If we haven't reached the end of the list of error details, add a newline separator between each error
+			if index < len(e.Details)-1 {
+				response += "\n"
 			}
 		}
+
 		return response
 	}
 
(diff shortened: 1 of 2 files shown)


client/errors.go
```diff
@@ -18,24 +18,59 @@ type DetailedError struct {
...
+	}
+
+	if td.Description != "" {
+		response += td.Description
+	}
+
+	return response
+}
+
+func (e DetailedError) detailsToString() string {
```
lafentres, line 55: i need to move this back into the error function, i don't think it makes sense to move it out

jharley: Agree! Please do and we'll get this landed ✨

client/errors.go
```diff
@@ -18,24 +18,59 @@ type DetailedError struct {
...
 	// Details is an array of structured objects that give details about the error.
-	Details []struct {
-		Code        string `json:"code"`
-		Description string `json:"description"`
-		Field       string `json:"field"`
-	} `json:"type_detail,omitempty"`
+	Details []TypeDetail `json:"type_detail,omitempty"`
+}
+
+type TypeDetail struct {
```
jharley, line 24: Maaaybe worth calling this `ErrorTypeDetail` or some thing? Naming is hard 🤷🏻

jharley (changes requested):
Thanks for catching this and writing such a lovely fix! Agree `Detailederror.detailsToString` should move back into `Error` but otherwise this is great

client/errors.go
lafentres: done!

client/errors.go
lafentres: done!

jharley (approved):
🌮 ✨

PR #393, opened 2023-11-14
client: better support for nondefault api hosts

cewkrupa:
This PR updates the client to support non-default API hosts by looking at the `HONEYCOMB_API_URL` environment variable.

Additionally, it:
- Brings CONTRIBUTING.md up to speed with the current state of the tests
- Updates `scripts/setup-testsuite-dataset` with all the columns the tests need

CI tests should still pass as expected.


CONTRIBUTING.md
```diff
@@ -24,11 +24,26 @@ Hashicorp has a tool to preview documentation. Visit [registry.terraform.io/tool
...
-Next, **initialize the dataset**. The helper script [setup-testsuite-dataset](scripts/setup-testsuite-dataset) will create the dataset and required columns that are used in the tests.
+Additionally, the test for a Slack recipient requires that the Slack authorization be [set up with the team ahead of time](https://docs.honeycomb.io/working-with-your-data/triggers/#slack)
+
+First, **create an API key**. This key should have the following permissions:
+- Create Datasets
+- Manage Queries and Columns
+- Manage Public Boards
+- Manage Triggers
+- Manage Recipients
+- Manage Markers
```
jharley, line 39: Worth mentioning the other two permissions for SLOs and Query Data results?

CONTRIBUTING.md
```diff
@@ -24,11 +24,26 @@ Hashicorp has a tool to preview documentation. Visit [registry.terraform.io/tool
...
+
+
+Next, **initialize the dataset**. The helper script [setup-testsuite-dataset](scripts/setup-testsuite-dataset) will create the dataset and required columns that are used in the tests. You will need to use Bash 4 to run this script.
```
jharley, line 42:
```suggestion
Next, **initialize the dataset**. The helper script [setup-testsuite-dataset](scripts/setup-testsuite-dataset) will create the dataset and required columns that are used in the tests. You will need to use Bash 4+ to run this script.
```

.tool-versions
```diff
@@ -0,0 +1 @@
+golang 1.20.8
```
jharley, line 1: FWIW if you add `legacy_version_file = yes` to `~/.asdfrc` it will use the current one. Happy to have both 👍🏻

internal/provider/burn_alert_resource_test.go
```diff
@@ -65,6 +66,13 @@ func TestAcc_BurnAlertResourceUpgradeFromVersion015(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					testAccEnsureBurnAlertExists(t, "honeycombio_burn_alert.test"),
 				),
+				SkipFunc: func() (bool, error) {
```
jharley, line 69: maybe worth a comment to explain why we're skipping this?

internal/provider/trigger_resource_test.go
```diff
@@ -68,6 +69,13 @@ func TestAcc_TriggerResourceUpgradeFromVersion014(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					testAccEnsureTriggerExists(t, "honeycombio_trigger.test"),
 				),
+				SkipFunc: func() (bool, error) {
```
jharley, line 72: As above

scripts/setup-testsuite-dataset
jharley, line 1: Thanks for finding these!

jharley (approved):
Few comments, all non-blocking

.tool-versions
cewkrupa: 🤦 dunno how I missed this, but I already had `legacy_version_file` set and asft seems to be picking it up just fine. I'll remove this file so we're not adding extra junk

PR #398, opened 2023-11-16
fix(slo, burn_alert, client) missing resources shouldn't error but be recreated

jharley:
v0.19.0's client re-write introduced new client error handling but mishandled the classic "re-create when missing remotely" behaviour that Terraform in known -- and loved -- for.

The TL;DR for the fix is that pointers to errors and `errors.As()` are not a good mix for successful error checking.

Initially reported as a bug via Honeycomb Support.

### DIFF
--- a/internal/provider/burn_alert_resource.go
+++ b/internal/provider/burn_alert_resource.go
@@ -195,8 +195,6 @@ func (r *burnAlertResource) ValidateConfig(ctx context.Context, req resource.Val
 	// When alert_type is budget_rate, check that budget rate
 	// attributes are configured and exhaustion_minutes is not configured
 	validateAttributesWhenAlertTypeIsBudgetRate(data, resp)
-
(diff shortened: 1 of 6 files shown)


jharley:
Currently missing a test to prove this out automatically, but verified with manual testing.

internal/provider/burn_alert_resource.go
```diff
@@ -195,8 +195,6 @@ func (r *burnAlertResource) ValidateConfig(ctx context.Context, req resource.Val
 	// When alert_type is budget_rate, check that budget rate
 	// attributes are configured and exhaustion_minutes is not configured
 	validateAttributesWhenAlertTypeIsBudgetRate(data, resp)
-
-	return
```
jharley, line 199: Linter was complaining about this

brookesargent (approved):

PR #401, opened 2023-11-20
chore: first pass at moving away from pre-created columns

jharley:
First step toward removing the dependency on `scripts/setup-testsuite-dataset` and also not being in a position that some failed test runs cause subsequent runs to fail due to conflicts.

In addition to randomly generating strings, the idea is to prefix the things we create with `test.` and make email addresses in the `@example.com` domain so that we can in future make use of [Sweepers](https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests/sweepers) to remove anything starting with `test.` or in the `@example.com` domain.

More to improve here, but this was intended to be a "time boxed" first pass.

### DIFF
--- a/client/dataset_definitions_test.go
+++ b/client/dataset_definitions_test.go
@@ -4,28 +4,60 @@ import (
 	"context"
 	"testing"
 
+	"github.com/honeycombio/terraform-provider-honeycombio/internal/helper/test"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestDatasetDefinitions(t *testing.T) {
+	t.Parallel()
+
 	ctx := context.Background()
 
 	c := newTestClient(t)
 	dataset := testDataset(t)
+	definitionDefaults := DatasetDefinitionDefaults()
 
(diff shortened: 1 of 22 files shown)


client/dataset_definitions_test.go
```diff
@@ -4,28 +4,60 @@ import (
...
 func TestDatasetDefinitions(t *testing.T) {
+	t.Parallel()
+
 	ctx := context.Background()
 
 	c := newTestClient(t)
 	dataset := testDataset(t)
+	definitionDefaults := DatasetDefinitionDefaults()
+
+	// ensure default definition columns exist -- create any which may be missing.
```
cewkrupa, line 21: Does this mean we can remove some or all of required columns section from the setup script? https://github.com/honeycombio/terraform-provider-honeycombio/blob/main/scripts/setup-testsuite-dataset#L11

cewkrupa (approved):
Love these changes, I think they're gonna make testing much easier!

client/dataset_definitions_test.go
jharley: We're not quite there yet, I don't think. I need to do a deeper pass in the provider code to make sure we're not dependant on any of them.

PR #408, opened 2023-12-12
Fix description of end_time in trigger docs

alouie-sfdc:
Accuracy of documentation.

This change fixes a copy/paste error in the docs for the `end_time` of an `evaluation_schedule`.

It only affects the documentation. A review should be all that's needed.


brookesargent (approved):
Thanks, @alouie-sfdc!

PR #409, opened 2023-12-13
docs: cleanup value_string reference

brookesargent:
Updates the note about `in` and `not-in` to reference `value` instead of `value_string` (which is deprecated) as well as a couple other cleanups related to the move to `value`


jharley (approved):
✨ 📄

PR #412, opened 2023-12-14
fix: legacy api key typo

### LINKED ISSUE #410: Legacy `HONEYCOMBIO_APIKEY` environment variable doesn't work anymore
**Versions**
It works in `0.16.0` but doesn't work in `0.20.2`. I haven't checked the versions in between.

**Steps to reproduce**

1. Set `HONEYCOMBIO_APIKEY` to your Honeycomb API key, but don't set `HONEYCOMB_API_KEY`
2. You'll get an error when trying to use the provider.

**Additional context**
The [docs](https://registry.terraform.io/providers/honeycombio/honeycombio/latest/docs#authentication) specify that the legacy environment variable is `HONEYCOMBIO_APIKEY`, but I believe there's a bug in the code here: https://github.com/honeycombio/terraform-provider-honeycombio/blob/24f9781240a51edef559707cb14c1a7fab7390de/client/client.go#L29

The last `_` shouldn't be present.

The workaround is to use the preferred `HONEYCOMB_API_KEY`.

brookesargent:

- Closes #410 

Removes extra underscore in the legacy API key so it will start working again

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -26,7 +26,7 @@ const (
 	DefaultAPIHostEnv = "HONEYCOMB_API_HOST"
 	DefaultAPIKeyEnv  = "HONEYCOMB_API_KEY"
 	// Deprecated: use DefaultAPIKeyEnv instead. To be removed in v1.0
-	LegacyAPIKeyEnv  = "HONEYCOMBIO_API_KEY"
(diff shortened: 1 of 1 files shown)


jharley (approved):
🐛✨

PR #416, opened 2023-12-21
refactor(client) - testsuite and constructor improvements

jharley:
This moves the embedded client's test suite into the package `client_test` and enforces a fully external test of the client. A few functional changes needed to happen to make this possible, but it felt like a cleaner way to test the client given it's intended purpose.

A breaking change is included in this refactor: `client.NewClient` now takes no arguments and creates a client with the default configuration values, while `client.NewClientWithConfig` has been added to allow passing a custom configuration.

This felt a bit more standard (and less awkward), and will only affect other projects pulling in the client package.


client/client.go
```diff
@@ -89,8 +90,14 @@ func DefaultConfig() *Config {
...
-// NewClient creates a new Honeycomb API client.
-func NewClient(config *Config) (*Client, error) {
+// NewClient creates a new Honeycomb API client with default settings.
+func NewClient() (*Client, error) {
+	return NewClientWithConfig(DefaultConfig())
+}
+
+// NewClientWithConfig creates a new Honeycomb API client using the provided
+// Config.
+func NewClientWithConfig(config *Config) (*Client, error) {
```
cewkrupa, line 100: Great call to add a client constructor that takes a config!

cewkrupa (approved):
Looks fantastic!!

PR #433, opened 2024-02-15
fix(recipients): better handle the case where a recipient is modified outside of Terraform

jharley:
# Summary

If a recipient was updated outside of Terraform (i.e. the UI or via the API in some other way) the provider would get rather cranky and spit out an error like:

> Could not find Recipient xysef83ryH in state

This adjusts that behaviour to instead accept that the recipient is defined remotely and allow the rest of Terraform's diffing process to handle the difference and present that plan to the user.

Effects `r/burn_alert` and `r/trigger`

---
  - https://app.asana.com/0/0/1206583960682330

### DIFF
--- a/internal/provider/notification_recipients.go
+++ b/internal/provider/notification_recipients.go
@@ -85,14 +85,15 @@ func notificationRecipientSchema(allowedTypes []client.RecipientType) schema.Set
 	}
 }
 
-func reconcileNotificationRecipientState(remote []client.NotificationRecipient, state []models.NotificationRecipientModel) []models.NotificationRecipientModel {
+func reconcileReadNotificationRecipientState(remote []client.NotificationRecipient, state []models.NotificationRecipientModel) []models.NotificationRecipientModel {
 	if state == nil {
 		// if we don't have any state, we can't reconcile anything so just return the remote recipients
 		return flattenNotificationRecipients(remote)
 	}
 
 	recipients := make([]models.NotificationRecipientModel, len(remote))
-	// match the remote recipients to those in the state sorting out type+target vs ID
+	// match the remote recipients to those in the state
+	// in an effort to preserve the id vs type+target distinction
 	for i, r := range remote {
 		idx := slices.IndexFunc(state, func(s models.NotificationRecipientModel) bool {
 			if !s.ID.IsNull() {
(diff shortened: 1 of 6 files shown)


brookesargent (approved):
Nice tests and appreciate the comments in `reconcileNotificationRecipientState` to make it clear what's going on! ✨

internal/provider/notification_recipients.go
```diff
@@ -84,6 +85,33 @@ func notificationRecipientSchema(allowedTypes []client.RecipientType) schema.Set
...
+				return s.ID.ValueString() == r.ID
+			}
+			return s.Type.ValueString() == string(r.Type) && s.Target.ValueString() == r.Target
+		})
+		if idx < 0 {
+			// if we didn't find a match, use the recipient as specified in remote
+			recipients[i] = notificationRecipientToModel(r)
+		} else {
+			// if we found a match, use the stored recipient
+			recipients[i] = state[idx]
```
dstrelau, line 108: Are `state` and `remote` guaranteed to be of the same length?

internal/provider/notification_recipients.go
```diff
@@ -106,19 +134,23 @@ func expandNotificationRecipients(n []models.NotificationRecipientModel) []clien
...
+
+func notificationRecipientToModel(r client.NotificationRecipient) models.NotificationRecipientModel {
+	rcpt := models.NotificationRecipientModel{
+		ID:     types.StringValue(r.ID),
+		Type:   types.StringValue(string(r.Type)),
+		Target: types.StringValue(r.Target),
+	}
+	if r.Details != nil {
+		rcpt.Details = make([]models.NotificationRecipientDetailsModel, 1)
+		rcpt.Details[0].PDSeverity = types.StringValue(string(r.Details.PDSeverity))
```
dstrelau, line 152:
```suggestion
		rcpt.Details = []models.NotificationRecipientDetailsModel{
			{PDSeverity: types.StringValue(string(r.Details.PDSeverity))},
		}
```

internal/provider/notification_recipients.go
jharley: Ah, good catch! They are not: I'll put in a check

dstrelau: I think I was misreading index variables here. This is fine 👍

dstrelau (approved):

PR #437, opened 2024-02-16
fix: handle column and derived column imports with slashes in the name

jharley:
Some columns and derived columns have slashes in the name. Prior to this change, trying to import either with slashes in the name would result in a message like:

```shell
$ terraform import 'honeycombio_column.test' 'test/my/column/name'
honeycombio_column.test: Importing from ID "test/my/column/name"...
honeycombio_column.test: Import prepared!
  Prepared honeycombio_column for import
honeycombio_column.test: Refreshing state... [id=test/my/column/name]
╷
│ Error: Cannot import non-existent remote object
│
│ While attempting to import an existing object to "honeycombio_column.test", the provider detected that no object exists with the given id. Only pre-existing objects can be imported; check that the id is correct
│ and that it is associated with the provider's configured region or endpoint, or use "terraform apply" to create a new remote object for this resource.
╵
```

Most (all?) other resources are imported by ID, but this subtle change in splitting does the trick here.

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -61,6 +61,9 @@ func TestClient_InvalidConfig(t *testing.T) {
 func TestClient_IsClassic(t *testing.T) {
 	t.Parallel()
 
+	// load environment values from a .env, if available
+	_ = godotenv.Load("../.env")
+
 	ctx := context.Background()
 	apiKey, ok := os.LookupEnv(client.DefaultAPIKeyEnv)
 	if !ok {
--- a/honeycombio/resource_column.go
+++ b/honeycombio/resource_column.go
@@ -3,7 +3,6 @@ package honeycombio
 import (
 	"context"
 	"errors"
-	"fmt"
 	"strings"
 	"time"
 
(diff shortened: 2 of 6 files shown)


client/client_test.go
```diff
@@ -61,6 +61,9 @@ func TestClient_InvalidConfig(t *testing.T) {
 func TestClient_IsClassic(t *testing.T) {
 	t.Parallel()
 
+	// load environment values from a .env, if available
+	_ = godotenv.Load("../.env")
```
jharley, line 65: unrelated, but snuck this in as the test was broken locally

honeycombio/resource_column.go
```diff
@@ -88,14 +87,13 @@ func newColumn() *schema.Resource {
...
-		return nil, fmt.Errorf("invalid import ID, supplied ID must be written as <dataset>/<column name>")
+	importID := d.Id()
+	idx := strings.Index(importID, "/")
+	if idx < 0 {
+		return nil, errors.New("invalid import ID, supplied ID must be written as <dataset>/<column name>")
 	}
-
-	dataset := strings.Join(idSegments[0:len(idSegments)-1], "/")
-	name := idSegments[len(idSegments)-1]
+	dataset := importID[:idx]
```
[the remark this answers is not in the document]
jharley: It was never accurate: we need the slug which can’t have a `/`

cewkrupa (approved):

honeycombio/resource_column.go
```diff
@@ -88,14 +87,13 @@ func newColumn() *schema.Resource {
 
 func resourceColumnImport(ctx context.Context, d *schema.ResourceData, i interface{}) ([]*schema.ResourceData, error) {
 	// import ID is of the format <dataset>/<column name>
-	// note that the dataset name can also contain '/'
-	idSegments := strings.Split(d.Id(), "/")
-	if len(idSegments) < 2 {
-		return nil, fmt.Errorf("invalid import ID, supplied ID must be written as <dataset>/<column name>")
+	importID := d.Id()
+	idx := strings.Index(importID, "/")
```
dstrelau, line 91: You might consider `strings.Cut`

dstrelau (approved):

honeycombio/resource_column.go
jharley: Thanks! Much nicer

PR #439, opened 2024-02-21
fix: query specification - always treat 'in','not-in' filter values as an array

jharley:
`in`, and `not-in` filter values should always be treated as an array. This allows for the possibility of filters like `app.mything.id not-in 38374`

### DIFF
--- a/honeycombio/data_source_query_specification.go
+++ b/honeycombio/data_source_query_specification.go
@@ -355,7 +355,8 @@ func extractFilter(d *schema.ResourceData, index int) (honeycombio.FilterSpec, e
 	valueSet := false
 	if v, ok := d.GetOk(fmt.Sprintf("filter.%d.value", index)); ok {
 		if filter.Op == honeycombio.FilterOpIn || filter.Op == honeycombio.FilterOpNotIn {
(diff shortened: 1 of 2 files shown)


honeycombio/data_source_query_specification.go
```diff
@@ -353,49 +353,55 @@ func extractFilter(d *schema.ResourceData, index int) (honeycombio.FilterSpec, e
...
-	if vbOk {
+	if v, ok := d.GetOk(fmt.Sprintf("filter.%d.value_boolean", index)); ok {
 		if valueSet {
 			return filter, fmt.Errorf(multipleValuesError)
 		}
-		filter.Value = vb
+		filter.Value = v
 	}
 
 	if filter.Op == honeycombio.FilterOpIn || filter.Op == honeycombio.FilterOpNotIn {
```
jharley, line 393: I understand we're doing "it's an in,not-in" filter things in two places at the moment, but this second check is a function of the deprecated `value_*` options. Once they're removed we can do this all in one place above

cewkrupa (approved):
🎉 looks great!

PR #454, opened 2024-03-18
feat: migrate r/query to Framework

jharley:
This migrates `r/query` from the Plugin SDKv2 to the Plugin Framework and introduces import support for queries to support [generating configuration on import](https://developer.hashicorp.com/terraform/language/import/generating-configuration).

As a bonus, found a bug in query equivalency with the default `limit` of `1000`.


honeycombio/data_source_column_test.go
```diff
@@ -44,8 +44,8 @@ func TestAccDataSourceHoneycombioColumn_basic(t *testing.T) {
 	})
 
 	resource.Test(t, resource.TestCase{
-		PreCheck:          testAccPreCheck(t),
-		ProviderFactories: testAccProviderFactories,
+		PreCheck:                 testAccPreCheck(t),
+		ProtoV5ProviderFactories: testAccProtoV5ProviderFactory,
```
jharley, line 48: All of these updates in the SDK-based tests were needed eventually as they need to be given a provider factory which can access the Framework-based resources/datasources/

brookesargent (approved):

PR #461, opened 2024-04-18
ci: add eu integration test flow

brookesargent:

This duplicates the existing integration test setup to run against our newer EU infrastructure.

I created a Terraform Provider team in the EU region and configured the testacc setup and tests in the Action to take the honeycomb api endpoint as an env var.

You can see the integration tests successfully ran in US and EU on this PR.

---
  - https://app.asana.com/0/0/1205935108176450

### DIFF
--- a/scripts/setup-testsuite-dataset
+++ b/scripts/setup-testsuite-dataset
@@ -3,7 +3,6 @@
 # the integration testsuite requires that a dataset and particular columns already exist
 # this little script handles that bit of bootstrapping in an idempotent fashion
 
-HONEYCOMB_API="${HONEYCOMB_API:=https://api.honeycomb.io}"
 COLUMNS_API="${HONEYCOMB_API}/1/columns"
 DATASETS_API="${HONEYCOMB_API}/1/datasets"
 
--- a/scripts/setup-testsuite-dataset
+++ b/scripts/setup-testsuite-dataset
@@ -2,7 +2,7 @@
 
 # the integration testsuite requires that a dataset and particular columns already exist
 # this little script handles that bit of bootstrapping in an idempotent fashion
-echo "$HONEYCOMB_API"
+
 COLUMNS_API="${HONEYCOMB_API}/1/columns"
 DATASETS_API="${HONEYCOMB_API}/1/datasets"
 
--- a/scripts/setup-testsuite-dataset
+++ b/scripts/setup-testsuite-dataset
@@ -3,8 +3,8 @@
 # the integration testsuite requires that a dataset and particular columns already exist
 # this little script handles that bit of bootstrapping in an idempotent fashion
 
-COLUMNS_API="${HONEYCOMB_API}/1/columns"
-DATASETS_API="${HONEYCOMB_API}/1/datasets"
+COLUMNS_API="${HONEYCOMB_API_ENDPOINT}/1/columns"
+DATASETS_API="${HONEYCOMB_API_ENDPOINT}/1/datasets"
 
 # a map of required columns for the integration tests and their types
 declare -A REQUIRED_COLUMNS=(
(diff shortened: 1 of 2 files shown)


scripts/setup-testsuite-dataset
```diff
@@ -3,9 +3,8 @@
 # the integration testsuite requires that a dataset and particular columns already exist
 # this little script handles that bit of bootstrapping in an idempotent fashion
 
-HONEYCOMB_API="${HONEYCOMB_API:=https://api.honeycomb.io}"
-COLUMNS_API="${HONEYCOMB_API}/1/columns"
-DATASETS_API="${HONEYCOMB_API}/1/datasets"
+COLUMNS_API="${HONEYCOMB_API_ENDPOINT}/1/columns"
```
jharley, line 6: I think doing the default message below means that `COLUMNS_API` and `DATASETS_API` will get set to `/1/columns` because the variable is missing at the time of them being set.

Maybe easier to just do it this way? If you want to bring the nicer "using the default" style message up to the top that's great too :)

```suggestion
HONEYCOMB_API_ENDPOINT="${HONEYCOMB_API_ENDPOINT:=https://api.honeycomb.io}"
COLUMNS_API="${HONEYCOMB_API_ENDPOINT}/1/columns"
```

scripts/setup-testsuite-dataset
```diff
@@ -82,6 +81,11 @@ if [ -z "${HONEYCOMB_DATASET+x}" ]; then
   exit 255
 fi
 
+if [ -z "${HONEYCOMB_API_ENDPOINT+x}" ]; then
```
jharley, line 84: I think setting this here is too late as `COLUMNS_API` and `DATASETS_API` will already be set with the empty string?

jharley (commented):
Thanks for this! Looking great.

Left a comment about the (hacky, and very much needs to die 😢) bash script

scripts/setup-testsuite-dataset
jharley: from the "test us" output:

```
HONEYCOMB_API_ENDPOINT is not set, using default https://api.honeycomb.io/
2024-04-18T14:27:20+0000 error creating dataset "testacc"
2024-04-18T14:27:20+0000 created dataset "testacc}"
2024-04-18T14:27:20+0000 error creating column "trace.trace_id"
2024-04-18T14:27:20+0000 created column "trace.trace_id" in "testacc"
2024-04-18T14:27:20+0000 error creating column "app.tenant"
2024-04-18T14:27:20+0000 created column "app.tenant" in "testacc"
2024-04-18T14:27:20+0000 error creating column "trace.span_id"
2024-04-18T14:27:20+0000 created column "trace.span_id" in "testacc"
2024-04-18T14:27:20+0000 error creating column "duration_ms"
2024-04-18T14:27:20+0000 created column "duration_ms" in "testacc"
2024-04-18T14:27:20+0000 error creating column "trace.parent_id"
2024-04-18T14:27:20+0000 created column "trace.parent_id" in "testacc"
2024-04-18T14:27:20+0000 error creating column "app.error"
2024-04-18T14:27:20+0000 created column "app.error" in "testacc"
2024-04-18T14:27:20+0000 error creating column "service_name"
2024-04-18T14:27:20+0000 created column "service_name" in "testacc"
2024-04-18T14:27:20+0000 error creating column "column_2"
2024-04-18T14:27:20+0000 created column "column_2" in "testacc"
2024-04-18T14:27:20+0000 error creating column "column_1"
2024-04-18T14:27:20+0000 created column "column_1" in "testacc"
2024-04-18T14:27:20+0000 error creating column "name"
2024-04-18T14:27:20+0000 created column "name" in "testacc"
```

Which.. also means that script failing isn't failing the run, I guess :|

scripts/setup-testsuite-dataset
brookesargent: Ah whoops! I think I didn't realize the `:=` syntax is an if unset type of thing but just looked it up and putting it back like this seems like the simplest/best thing.

scripts/setup-testsuite-dataset
brookesargent: 🙃 that's what I get for assuming green = good and not poking further

jharley: It _should_ have caused "red = bad" 🙃 

We can fix that outside of this, tho'

jharley (approved):

PR #487, opened 2024-06-14
r/trigger: re-introduce query_json for deeper validation

jharley:
In [v0.2.0](https://github.com/honeycombio/terraform-provider-honeycombio/blob/main/CHANGELOG.md#020-jan-27-2022) we dropped `query_json` from `r/trigger` to smooth out few a gnarly bugs associated with the Plugin SDK. With that change the ability to more deeply validate that a Trigger's query met the requirements was sadly lost.

Since then, `r/trigger` has been re-written in the new Plugin Framework and with it we got the ability to determine if the configuration was building via Query ID or Query JSON. This change takes advantage of that ability and re-introduces deeper validation for Trigger query's so long as you use `query_json` in place of `query_id`

### DIFF
--- a/internal/provider/trigger_resource.go
+++ b/internal/provider/trigger_resource.go
@@ -2,6 +2,7 @@ package provider
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"regexp"
 	"strings"
@@ -9,6 +10,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework-validators/int64validator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/listvalidator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/booldefault"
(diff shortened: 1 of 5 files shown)


internal/provider/trigger_resource.go
```diff
@@ -385,10 +489,107 @@ func (r *triggerResource) ImportState(ctx context.Context, req resource.ImportSt
...
+	var data models.TriggerResourceModel
+	resp.Diagnostics.Append(req.Config.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// exit early if we don't have QueryJSON
+	if data.QueryJson.IsNull() || data.QueryJson.IsUnknown() {
+		return
+	}
```
cewkrupa, line 508: Does returning early here mean it successfully validated or that it failed?

jharley: Neither: we can't validate anything that isn't there

cewkrupa: Ok, maybe this is a mental model fail on my part -- I was thinking about the line above

```go
if resp.Diagnostics.HasError() {
		return
	}
```
which to me read a bit like "if there's an existing error the config of the trigger resource fails validation". But maybe it means something closer to "if there's an existing error [or we don't have queryJSON]", don't run further validation"? If that's right, that makes sense 👍

jharley: That's a different error case: it means we were unable to get the configuration so we fail out right away.

These functions don't have returned results (no boolean or error), but instead add to a `*resource.ValidateConfigResponse` -- often with the intention of adding more than one error or warning -- but there are also cases where you know it's not safe to continue.

cewkrupa (approved):

PR #492, opened 2024-06-18
feat: validate examples

### LINKED ISSUE #104: CI: apply examples to ensure they're valid
Ensure all provided examples (in `example/`) are valid and apply correctly. Perhaps post-testsuite success?

jharley:
Adds a light example validation workflow to ensure things are more or less correct. A full apply would be nice but involves a lot of setup and this is for sure "better than nothing".

Bonus: fixes trigger examples, and removes no-longer-used asana workflows.

- Closes #104


.github/workflows/validate-examples.yml
```diff
@@ -0,0 +1,28 @@
...
+              uses: actions/checkout@v2
+
+            - name: Setup Go
+              uses: actions/setup-go@v5
+              with:
+                  go-version-file: "go.mod"
+                  cache: true
+
+            - name: Build
+              run: go build -o tmp/providers/terraform-provider-honeycombio
```
jharley, line 22: This ensures we're running validation against what is about to land on main

mjayaram (approved):

PR #493, opened 2024-06-20
chore: d/query_specification - migrate to Plugin Framework

jharley:
The first of two PRs to make improvements to the `query_specification` datasource's `filter` block allowing more complex `value` options.

This is a migration of the datasource from the Plugin SDK to the Plugin Framework with as few changes as possible. However, I did decide to (finally?) drop the `value_*` options to `filter` as they have been deprecated for over two years now.

The test file for the datasource was kept as in-tact as possible to attempt to minimize the review effort.

The follow-on PR to this will convert the `filter` block's `value` to a [DynamicAttribute](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/dynamic-data#considerations).


internal/provider/slos_data_source_test.go
```diff
@@ -111,7 +111,6 @@ data "honeycombio_slos" "exact" {
 }
 `, dataset, testPrefix),
 				Check: resource.ComposeAggregateTestCheckFunc(
-					resource.TestCheckResourceAttr("data.honeycombio_slos.all", "ids.#", "5"),
```
jharley, line 114: This test had been flakey since the Boards testsuite starting making additional SLOs. 

I've just removed it for the time being as the other tests do a decent job ensuring the datasource is working.

internal/provider/query_specification_data_source_test.go
jharley, line 1: Most changes in this file are updates to the error message checks as the Framework allows more specific, attribute-level error reporting.

mjayaram (approved):

Issue #446, opened 2024-02-29
Add Sweepers for test cleanup


jharley:
The test suite is still leaving many things behind after a test run: notably Recipients even after a fully successful run. This gets extra sticky if a test run is unable to clean up (panic, cancelled early) and dangling resources end up causing conflicts on the following test runs.

Implementing [Sweepers](https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests/sweepers) may help ease the pain here, especially if the entire test suite ends up prefixing resources with `test-` or `test.` or some such thing.

PR #513, opened 2024-07-29
chore: limit integration test concurrency

jharley:
Limit integration test runs with [concurrency groups](https://docs.github.com/en/actions/using-jobs/using-concurrency).

Not currently using `cancel-in-progress` until [sweepers](https://github.com/honeycombio/terraform-provider-honeycombio/issues/446) are implemented and it's confirmed that they can run when receiving a cancellation signal.


cewkrupa (approved):
✨

PR #514, opened 2024-07-29
fix: allow auto ('0') granularity

### LINKED ISSUE #512: d/query_specification: failing validation on "auto" granularity = 0
Reported via Pollinators:

> with the [0.24.0] release we faced the following issue:
Error: invalid granularity
│ 
│   with data.honeycombio_query_specification.xxxx,
│   on xxxx.tf line 132, in data "xxxx" "xxxx":
│  132:   granularity = 0
│ 
│ granularity can not be less than time_range/1000

> For one this does not affect queries using the JSON based definition (which is kinda expected - I'd guess)
> Also: iirc granularity = 0  used to mean auto hence the value should be fine; For the time being we removed all granularity = 0 statements, relying on the implicit default, but I'm wondering if this is intentional.

jharley:
Allow 'auto' granularity to be specified as `0`

- Closes #512

### DIFF
--- a/client/v2/client.go
+++ b/client/v2/client.go
@@ -208,7 +208,7 @@ func (c *Client) retryHTTPCheck(
 	case http.StatusTooManyRequests:
 		// TODO: use new retry header timestamps to determine when to retry
 		return true, nil
-	case http.StatusBadGateway, http.StatusGatewayTimeout:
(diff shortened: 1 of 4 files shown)


jharley:
Also sneaks in a one liner to have the provider V2 client retry on HTTP 500s

brookesargent (approved):

PR #516, opened 2024-07-30
chore: add recipients sweeper

### LINKED ISSUE #446: Add Sweepers for test cleanup

jharley:
Adds [Sweepers](https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests/sweepers) to the project with an initial sweeper for recipients (which seem to cause the most test failures/conflicts).

This also attempts to get all recipients named with a `test.` prefix (`#test.` for Slack recipients) establishing the convention that if a thing's name starts with `test.` is can be "swept".

- Closes #446


dstrelau (approved):

.golangci.yml
```diff
@@ -35,7 +35,7 @@ linters:
         - unconvert
         - unparam
         - unused
-        - vet
+        - govet
```
jharley, line 38: `vet` linter is deprecated

PR #519, opened 2024-08-01
feat(client): V2 API client ratelimit improvements

### LINKED ISSUE #511: V2 API Client gracefully handles rate limits observing Rate-Limit headers
The V2 API Client currently "just retries" when it receives a 429.

The headers for V2 APIs contain detail on when to retry and the total quota which should be used to determine when to try again.

jharley:
This change teaches the V2 API client to respond to HTTP429s by backing off using the Ratelimit headers in the API's response.

- Closes #511

### DIFF
--- a/client/v2/limits.go
+++ b/client/v2/limits.go
@@ -1,0 +1,69 @@
+package v2
+
+import (
+	"fmt"
+	"math/rand"
+	"net/http"
+	"time"
+)
+
+const (
+	// HeaderRateLimit is the (draft07) recommended header from the IETF
+	// on rate limiting.
+	HeaderRateLimit = "RateLimit"
+	// The value of the header is formatted "limit=X, remaining=Y, reset=Z".
+	// Where:
+	//   - X is the maximum number of requests allowed in the window
+	//   - Y is the number of requests remaining in the window
+	//   - Z is the number of seconds until the limit resets
+	HeaderRateLimitFmt = "limit=%d, remaining=%d, reset=%d"
+
+	// HeaderRetryAfter is the RFC7231 header used to indicate when a client should
+	// retry requests in UTC time.
+	HeaderRetryAfter = "Retry-After"
+)
+
+var rnd *rand.Rand
+
+// init initializes the random number generator
+func init() {
+	rnd = rand.New(rand.NewSource(time.Now().UnixNano()))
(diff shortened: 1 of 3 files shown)


client/v2/limits.go
```diff
@@ -0,0 +1,75 @@
...
+	// HeaderRetryAfter is the RFC7231 header used to indicate when a client should
+	// retry requests in UTC time.
+	HeaderRetryAfter = "Retry-After"
+)
+
+var rng *rand.Rand
+
+func init() {
+	// initialize the random number generator
+	rng = rand.New(rand.NewSource(time.Now().UnixNano()))
```
dstrelau, line 30: Is there a reason not to use the rand package-level functions?

cewkrupa (approved):

client/v2/limits.go
```diff
@@ -0,0 +1,75 @@
...
+
+// parseRateLimitHeader parses the rate limit header into its constituent parts.
+//
+// The header is expected to be in the format "limit=X, remaining=Y, reset=Z".
+// Where:
+//   - X is the maximum number of requests allowed in the window
+//   - Y is the number of requests remaining in the window
+//   - Z is the number of seconds until the limit resets
+func parseRateLimitHeader(v string) (limit, remaining, reset int, err error) {
+	_, err = fmt.Sscanf(v, HeaderRateLimitFmt, &limit, &remaining, &reset)
```
dstrelau, line 73: I'm not sure if we're worried about the server changing the exact format of this header, but using `SScanf` like this means even an added or removed space is going to break things.

The actual [spec for SFV](https://datatracker.ietf.org/doc/html/rfc8941) is quite complex. There are seemingly [libraries](https://github.com/dunglas/httpsfv) to parse it generically. Maybe something like splitting on `,` and `=` would be slightly more robust without pulling in a whole library?

client/v2/limits.go
jharley: I honestly forgot `rand.Float64()` was there! Thanks!

client/v2/limits.go
jharley: AH! Thank you for connecting that to the standard for me.

I went ahead and used the library you linked anyway as it felt both proper and a good bit of learning.

PR #524, opened 2024-08-07
feat: add r/msteams_workflow_recipient

jharley:
Adds the new `r/msteams_workflow_recipient` resource and deprecates `r/msteams_recipient`.


docs/resources/msteams_recipient.md
```diff
@@ -2,6 +2,10 @@
 
 `honeycombio_msteams_recipient` allows you to define and manage an MSTeams recipient that can be used by Triggers or BurnAlerts notifications.
 
+-> **NOTE** Microsoft has deprecated Teams Incoming Webhooks.
```
brookesargent, line 5: would it be clearer to say something about "connectors" here?

brookesargent (approved):
left one comment on wording but lgtm provided the tests pass once we get the API logic in

docs/resources/msteams_recipient.md
jharley: Agreed! Will update.

PR #525, opened 2024-08-07
fix: environments data source test flake

jharley:
~The environments data source is the most recent victim of the "total count wrong" test flake. Before this, the SLO data source test had this before and it flaked so much I removed it #493.~

~This writes a helper to compare that the number of items in an attribute is _at least_ what we expect it to be and restores the SLO data source total count check.~

Remove the flakey all count comparison -- but continue to run a data source without a filter.


dstrelau:
I'm skeptical that this is a test worth keeping. "There's probably _some_ stuff in there" doesn't seem too useful to me. 🤔

jharley:
@dstrelau fair! I'll just rip out the test for `all` 👍🏻

dstrelau (approved):

PR #526, opened 2024-08-08
chore: improve V2 API error reporting

jharley:
Improves V2 API call error reporting/handling by making use of the JSON:API `source` field ([doc](https://jsonapi.org/format/#error-objects)) in error responses now being sent by the Honeycomb API.

Future work will skip converting these into the RFC7807 'Problem Detail' struct shared with the V1 client and convert them to Terraform diagnostics directly.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/google/go-querystring v1.1.0
 	github.com/hashicorp/go-cty v1.4.1-0.20200414143053-d3edf31b6320
 	github.com/hashicorp/go-retryablehttp v0.7.7
-	github.com/hashicorp/jsonapi v1.3.1
+	github.com/hashicorp/jsonapi v1.3.2-0.20240802183744-2490a9451c3d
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/google/go-querystring v1.1.0
 	github.com/hashicorp/go-cty v1.4.1-0.20200414143053-d3edf31b6320
 	github.com/hashicorp/go-retryablehttp v0.7.7
-	github.com/hashicorp/jsonapi v1.3.1
+	github.com/hashicorp/jsonapi v1.3.2-0.20240802183744-2490a9451c3d
```
jharley, line 12: This is the [commit](https://github.com/hashicorp/jsonapi/commit/2490a9451c3dc98a9f3ae4f9e317a0b8646fa7c9) with the `Source` addition ErrorObject.

Will switch to the next released version whenever it happens, but this seems fine in the interim

dstrelau (approved):

PR #530, opened 2024-08-13
feat: add ends/does-not-end-with

cartermp:
Companion to the upstream support we have now for this operator.


jharley (approved):
🚀 

thanks for this, @cartermp!

PR #535, opened 2024-08-22
feat: full lifecycle support for datasets

jharley:
# Summary

This brings full lifecycle management to Datasets: which is a fancy way of saying "datasets will be deleted on destroy now" with a side of "we will now fail on 'create' if the dataset already exists instead of silently importing the dataset" -- behaviour that previously existed with columns and derived columns and was removed because it was surprising and arguably dangerous.

This is a breaking change in behaviour, but one done in the name of being more Terraform-y and that feels "correct" as we move toward declaring the provider stable.

Datasets have deletion protection -- enabled by default -- which must be disabled to successfully destroy them, and matches the pattern used with Environments (`r/environment`) in v0.25.0.

# Breaking changes

* client: if the Datasets API returns an HTTP '200 OK' -- as opposed to '201 Created' -- on create the `Datasets.Create` method will now return `ErrDatasetExists`.
* r/dataset: will now do the more standard thing and error if a dataset with that name already exists in the Environment. The end-user can choose to adjust the name or import the existing dataset.

# Also included:

* a fly-by bug fix for v2 client calls not closing the HTTP response body
* a new data source for a single dataset (`d/dataset`) now that folks will potentially want to look them up without relying on the previous "import on exists" behaviour in `r/dataset`
* a re-write of the existing `d/datasets` in the Plugin Framework
  * this deprecates `starts_with` in favour of the standard `detail_filter` block
* a dataset 'sweeper' for cleaning up datasets having names beginning with `test.` in CI


internal/provider/dataset_resource.go
```diff
@@ -0,0 +1,292 @@
...
+		return
+	}
+	r.client = c
+}
+
+func (*datasetResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Manages a Dataset in a Honeycomb Environment.",
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
```
jharley, line 65: because we're still using the V5 provider protocol all resources (and datasources) need to have an ID.

In future we can drop this and the "id" can properly be the Slug

brookesargent (approved):
overall looking pretty good! appreciate the tests and docs notes on behavior prior to this PR 🙏

PR #536, opened 2024-08-23
rel: 0.27.0

jharley:


CHANGELOG.md
```diff
@@ -1,3 +1,32 @@
+# 0.27.0 (Aug 23, 2024)
+
+NOTES:
+
+**This release contains breaking and behavioural changes**
```
brookesargent, line 5: obviously both spellings are right in their own way, but not sure which is right in this context. I guess more of a general docs style guide question that provider docs would follow?
```suggestion
**This release contains breaking and behavioral changes**
```

CHANGELOG.md
```diff
@@ -1,3 +1,32 @@
...
+
+* `r/dataset`: will now attempt to *delete* datasets on destroy.
+  However, they come with delete protection (`delete_protection`) enabled by default and it must be disabled for the destroy to succeed.
+  This behaviour matches `r/envionment` as introduced in v0.25.0
```
brookesargent, line 9: same q as above

brookesargent, line 9:
```suggestion
  This behavior matches `r/environment` as introduced in v0.25.0
```

CHANGELOG.md
```diff
@@ -1,3 +1,32 @@
...
+* `r/dataset`: will now error on create if a Dataset with that name already exists in the Environment.
+  Previously, the dataset would be silently imported into state, but given deletion is now possible on destroy this behaviour has to changed to be more standard.
```
brookesargent, line 11: same q here too

CHANGELOG.md
```diff
@@ -1,3 +1,32 @@
...
+  The end-user can choose to either adjust the name or import the existing Dataset.
+* client: if the Datasets API returns an HTTP '200 OK' -- as opposed to '201 Created' -- on create the `Datasets.Create` method will now return `ErrDatasetExists`.
+
+ENHANCEMENTS:
+
+* *New Datasource*: `honeycombio_dataset` (#535)
```
brookesargent, line 17:
```suggestion
* *New Data Source*: `honeycombio_dataset` (#535)
```

CHANGELOG.md
jharley: Ha, my Canadian/British english (and spellchecker) is shining through. We use US spellings in our docs so I'll go with that: thanks for the catch!

CHANGELOG.md
jharley, line 11:
```suggestion
  Previously, the dataset would be silently imported into state, but given deletion is now possible on destroy this behavior has to changed to be more standard.
```

brookesargent (approved):

PR #538, opened 2024-09-03
chore(docs): standardize use of notes

jharley:
Noticed while browsing the provider docs that the use of "callouts" (largely used for notes) is a bit inconsistent.

This change attempts to fix that, while using the available yellow and red [callouts](https://developer.hashicorp.com/terraform/registry/providers/docs#callouts) for warnings and deprecation notices.

### DIFF
--- a/docs/data-sources/column.md
+++ b/docs/data-sources/column.md
@@ -2,7 +2,7 @@
 
 The `honeycombio_column` data source retrieves the details of a single column in a dataset.
 
--> **Note** Terraform will fail unless a column is returned by the search. Ensure that your search is specific enough to return a column.
+~> **Warning** Terraform will fail unless a column is returned by the search. Ensure that your search is specific enough to return a column.
 If you want to match multiple columns, use the `honeycombio_columns` data source instead.
 
 ## Example Usage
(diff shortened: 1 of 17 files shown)


docs/data-sources/column.md
```diff
@@ -2,7 +2,7 @@
 
 The `honeycombio_column` data source retrieves the details of a single column in a dataset.
 
--> **Note** Terraform will fail unless a column is returned by the search. Ensure that your search is specific enough to return a column.
+~> **Warning** Terraform will fail unless a column is returned by the search. Ensure that your search is specific enough to return a column.
```
cewkrupa, line 5: I think I'm still a lil confused what this will end up looking like based on this table. 

image

Will this end up as a yellow callout with the text prefix `**Note** **Warning**`, since it's using `~>`?

I also notice that it says `!>` is supposed to end up with a red callout and the `**Warning**` prefix,  but for all of our callouts that say "Warning" we're using `~>` -- is that what we want? (my guess is yes since they seem like less important warnings, but wanted to check)

jharley: Providing a `**Something**` overrides the default text prefix, so the above renders like this:

image

I should have linked their [Doc Preview Tool](https://registry.terraform.io/tools/doc-preview) which lets you try out their flavor of Markdown

cewkrupa: Got it, thanks! I'll bookmark that for later, I did not realize it was an override but that sounds perfect.

cewkrupa (approved):

PR #557, opened 2024-10-08
update regex and fix typos

brookesargent:

We fixed this in the API last week. Our regex previously allowed channel and DM ids but not user IDs which is important for a DM alert target that doesn't have a conversation ID yet. Also found a couple typos in CONTRIBUTING.md that are fixed here.


jharley (approved):
✨🧹

PR #577, opened 2024-11-15
chore: tidy up lint config

jharley:
`unparam` felt a bit extra in tests, so disabled it for test files and then cleaned up all the `nolint` directives

### DIFF
--- a/honeycombio/resource_slo.go
+++ b/honeycombio/resource_slo.go
@@ -86,12 +86,7 @@ func resourceSLOCreate(ctx context.Context, d *schema.ResourceData, meta interfa
 	}
 
 	dataset := d.Get("dataset").(string)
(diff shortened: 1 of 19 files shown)


honeycombio/resource_slo.go
```diff
@@ -86,12 +86,7 @@ func resourceSLOCreate(ctx context.Context, d *schema.ResourceData, meta interfa
 	}
 
 	dataset := d.Get("dataset").(string)
-	s, err := expandSLO(d)
```
jharley, line 89: `unparam` was correct to call out that this function could never return an error (the signature was tying to match a bunch of the other `expand*` helpers)

cewkrupa (approved):
🧹

PR #581, opened 2024-11-20
chore: include request id in debug logs and diagnostics

cewkrupa:

When calls to the Honeycomb API fail in tests and in real operation, we don't get much detail about what's wrong - just an error message if we're lucky. We've recently added a `Request-Id` header to our API responses to aid in debugging.

This PR pulls the `Request-Id` header of failed requests into the error messages, so they'll show up in test failures, in debug mode, and in regular operation.


client/client.go
```diff
@@ -141,7 +141,7 @@ func NewClientWithConfig(config *Config) (*Client, error) {
 		// if enabled we log all requests and responses to sterr
 		client.httpClient.Logger = log.New(os.Stderr, "", log.LstdFlags)
 		client.httpClient.ResponseLogHook = func(l retryablehttp.Logger, resp *http.Response) {
-			l.Printf("[DEBUG] Response: %s %s", resp.Request.Method, resp.Request.URL.String())
```
cewkrupa, line 144: I changed this to say "Request" because we weren't actually logging much about the response, just the request. I added the request ID and the status as well.

internal/helper/diag.go
```diff
@@ -53,8 +53,8 @@ func AddDiagnosticOnError(diag *diag.Diagnostics, summary string, err error) boo
 }
 
 func (d DetailedErrorDiagnostic) Detail() string {
+	response := fmt.Sprintf("ID: %s\n", d.e.ID)
```
cewkrupa, line 56: We're doing some confusing things with looping over the Details of the error to generate the message, and it's otherwise not super consistent as far as I can tell. Super open to other ideas of how to format this.

PR #582, opened 2024-11-21
fix: webhook recipient test

brookesargent:

We're seeing a 500 error that I believe is caused by the update in this test not actually updating the template. CI on this branch didn't cause the same 500 so I'd like to merge it in to avoid spamming ourselves with 500s when this test suite runs. I think there's still more to consider about why this is happening both on the TF side and the API side but this is a bandaid for the noise.


jharley (approved):
🩹 👍🏻

PR #583, opened 2024-11-22
feat: template variables

brookesargent:

Support webhook recipient template variables in Terraform

### DIFF
--- a/internal/provider/webhook_recipient_resource.go
+++ b/internal/provider/webhook_recipient_resource.go
@@ -128,7 +128,7 @@ func (*webhookRecipientResource) Schema(_ context.Context, _ resource.SchemaRequ
 							Required:    true,
 							Description: "The name of the variable",
 							Validators: []validator.String{
-								stringvalidator.LengthAtMost(63),
+								stringvalidator.LengthBetween(1, 64),
 								stringvalidator.RegexMatches(webhookTemplateNameRegex, "must be an alphanumeric string beginning with a lowercase letter"),
 							},
 						},
--- a/internal/provider/webhook_recipient_resource.go
+++ b/internal/provider/webhook_recipient_resource.go
@@ -3,7 +3,10 @@ package provider
 import (
 	"context"
 	"errors"
+	"fmt"
+	"regexp"
 
+	"github.com/hashicorp/terraform-plugin-framework-validators/setvalidator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
 	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
@@ -29,7 +32,8 @@ var (
 	_ resource.ResourceWithImportState    = &webhookRecipientResource{}
 	_ resource.ResourceWithValidateConfig = &webhookRecipientResource{}
 
-	webhookTemplateTypes = []string{"trigger", "exhaustion_time", "budget_rate"}
+	webhookTemplateTypes     = []string{"trigger", "exhaustion_time", "budget_rate"}
+	webhookTemplateNameRegex = regexp.MustCompile(`^[a-z](?:[a-zA-Z0-9]+$)?$`)
 )
 
 type webhookRecipientResource struct {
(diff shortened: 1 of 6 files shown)


internal/provider/webhook_recipient_resource_test.go
jharley, line 1: great test additions!

Can you please add a test for the case where the default isn't provided (either on creation or update -- maybe both?)

brookesargent (commented):
made some changes to the test file that are a bit annoying to read. I pulled the "errors when they should" `t.Run` out of the main test and into individual tests that have descriptive names about what exactly is being tested as the number of validations/tests was getting a bit unwieldy

internal/provider/webhook_recipient_resource.go
```diff
@@ -172,10 +203,16 @@ func (r *webhookRecipientResource) Create(ctx context.Context, req resource.Crea
 	} else {
 		state.Secret = types.StringNull()
 	}
+
+	// to prevent confusing if/else blocks, set null by default and override it if we have that detail on the recipient
```
jharley, line 207: ✨

internal/provider/webhook_recipient_resource.go
```diff
@@ -420,3 +500,32 @@ func validateAttributesWhenTemplatesIncluded(ctx context.Context, data models.We
 
 	}
 }
+
+func validateAttributesWhenVariablesIncluded(ctx context.Context, data models.WebhookRecipientModel, resp *resource.ValidateConfigResponse) {
+	var templates []models.WebhookTemplateModel
```
jharley, line 505: nit: given that we're now decoding templates in this function now as well it might be worth just moving the logic from both of these validate helpers right into `ValidateConfig`

internal/provider/webhook_recipient_resource_test.go
```diff
@@ -151,6 +173,232 @@ resource "honeycombio_webhook_recipient" "test" {
...
+			Steps: []resource.TestStep{
+				{
+					Config: fmt.Sprintf(`
+resource "honeycombio_webhook_recipient" "test" {
+  name = "%s"
+	url  = "%s"
+
+	variable {
+	  name = "variable1"
+      default_value = ""
```
jharley, line 315: Because `default_value` is optional it'd be good to test that case by leaving it out completely
```suggestion
```

internal/provider/webhook_recipient_resource_test.go
```diff
@@ -161,11 +409,160 @@ resource "honeycombio_webhook_recipient" "test" {
...
+
+			variable {
+			  name   = "severity"
+		      default_value = "critical"
+		    }
+		}`, name, url),
+					Destroy:     true,
+					ExpectError: regexp.MustCompile(`cannot configure a "variable" without also configuring a "template"`),
+				},
+				// final successful configuration to prevent dangling resource error
```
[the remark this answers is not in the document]
jharley: Oh, nice one! And, nice set of tests

internal/provider/webhook_recipient_resource.go
```diff
@@ -332,19 +383,37 @@ func webhookTemplatesToClientPayloads(ctx context.Context, set types.Set, diags
...
+
+		clientVars[i] = tmplVar
+	}
+	clientWebhookPayloads.TemplateVariables = clientVars
+
 	return clientWebhookPayloads
 }
 
-func clientPayloadsToWebhookTemplates(ctx context.Context, p *client.WebhookPayloads, diags *diag.Diagnostics) types.Set {
+func clientPayloadsToSets(ctx context.Context, p *client.WebhookPayloads, diags *diag.Diagnostics) (types.Set, types.Set) {
```
jharley, line 400: nit-ish: these functions are scoped to the whole package so a little more descriptive name would be good, I think
```suggestion
func clientPayloadsToWebhookTemplateSets(ctx context.Context, p *client.WebhookPayloads, diags *diag.Diagnostics) (types.Set, types.Set) {
```

jharley:
> made some changes to the test file that are a bit annoying to read. I pulled the "errors when they should" `t.Run` out of the main test and into individual tests that have descriptive names about what exactly is being tested as the number of validations/tests was getting a bit unwieldy

This looks great, thank you!

internal/provider/webhook_recipient_resource_test.go
brookesargent: ah makes sense, i missed the point 😅

jharley (approved):
Looks great, one note about a needed docs addition ✨

PR #586, opened 2024-11-27
feat: webhook headers

brookesargent:

Support headers on the webhook recipient resource.

### DIFF
--- a/client/recipient.go
+++ b/client/recipient.go
@@ -65,7 +65,8 @@ type RecipientDetails struct {
 	WebhookName string `json:"webhook_name,omitempty"`
 	WebhookURL  string `json:"webhook_url,omitempty"`
 	// webhook only
-	WebhookSecret string `json:"webhook_secret,omitempty"`
+	WebhookSecret  string          `json:"webhook_secret,omitempty"`
+	WebhookHeaders []WebhookHeader `json:"webhook_headers"`
 	// custom webhook
 	WebhookPayloads *WebhookPayloads `json:"webhook_payloads,omitempty"`
 }
@@ -92,6 +93,11 @@ type PayloadTemplate struct {
 type TemplateVariable struct {
 	Name    string `json:"name"`
 	Default string `json:"default_value"`
+}
+
+type WebhookHeader struct {
+	Key   string `json:"header"`
+	Value string `json:"value"`
 }
 
 // RecipientType holds all the possible recipient types.
--- a/client/recipient_test.go
+++ b/client/recipient_test.go
@@ -134,7 +134,7 @@ func TestRecipientsCustomWebhook(t *testing.T) {
 			require.Len(t, tr.Details.WebhookHeaders, 1)
 			require.Len(t, r.Details.WebhookHeaders, 1)
 			assert.Equal(t, tr.Details.WebhookHeaders[0].Key, r.Details.WebhookHeaders[0].Key)
-			assert.Equal(t, tr.Details.WebhookHeaders[0].Value, r.Details.WebhookHeaders[0].Key)
+			assert.Equal(t, tr.Details.WebhookHeaders[0].Value, r.Details.WebhookHeaders[0].Value)
 		})
 	}
 }
--- a/client/recipient_test.go
+++ b/client/recipient_test.go
@@ -131,10 +131,10 @@ func TestRecipientsCustomWebhook(t *testing.T) {
 			assert.Equal(t, tr.Details.WebhookSecret, r.Details.WebhookSecret)
 			assert.Equal(t, tr.Details.WebhookPayloads, r.Details.WebhookPayloads)
 			assert.Equal(t, tr.Details.WebhookPayloads.TemplateVariables, r.Details.WebhookPayloads.TemplateVariables)
-			require.Len(t, tr.Details.WebhookHeaders, 1)
-			require.Len(t, r.Details.WebhookHeaders, 1)
-			assert.Equal(t, tr.Details.WebhookHeaders[0].Key, r.Details.WebhookHeaders[0].Key)
-			assert.Equal(t, tr.Details.WebhookHeaders[0].Value, r.Details.WebhookHeaders[0].Value)
+			if assert.Len(t, r.Details.WebhookHeaders, 1) {
+				assert.Equal(t, tr.Details.WebhookHeaders[0].Key, r.Details.WebhookHeaders[0].Key)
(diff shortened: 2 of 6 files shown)


client/recipient.go
```diff
@@ -65,7 +65,8 @@ type RecipientDetails struct {
 	WebhookName string `json:"webhook_name,omitempty"`
 	WebhookURL  string `json:"webhook_url,omitempty"`
 	// webhook only
-	WebhookSecret string `json:"webhook_secret,omitempty"`
+	WebhookSecret  string          `json:"webhook_secret,omitempty"`
+	WebhookHeaders []WebhookHeader `json:"webhook_headers"`
```
brookesargent, line 69: I had omit empty set here since all the other ones do but it was causing an issue where len(0) when we want to unset all the headers wasn't getting sent and so the headers were always sticking around. lmk if there's a better way to do this in the TF schema instead

internal/provider/webhook_recipient_resource.go
```diff
@@ -145,6 +152,31 @@ func (*webhookRecipientResource) Schema(_ context.Context, _ resource.SchemaRequ
...
+				Validators: []validator.Set{
+					setvalidator.SizeAtMost(5),
+				},
+				NestedObject: schema.NestedBlockObject{
+					Attributes: map[string]schema.Attribute{
+						"name": schema.StringAttribute{
+							Required:    true,
+							Description: "The name or key for the header",
+							Validators: []validator.String{
+								stringvalidator.LengthBetween(1, 64),
```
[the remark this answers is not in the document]
cewkrupa: I just looked at the HTTP spec and it seems like there's no defined limit on how long header keys or values [can be](https://www.rfc-editor.org/rfc/rfc7230.html#section-3.2.5). I think this is probably a reasonable length to start for the name of the header; maybe we could double the size of the value to 512?

client/recipient.go
```diff
@@ -94,6 +95,11 @@ type TemplateVariable struct {
 	Default string `json:"default_value"`
 }
 
+type WebhookHeader struct {
+	Key   string `json:"header"`
```
cewkrupa, line 99:
```suggestion
	Name   string `json:"name"`
```
Should this be name instead of Key and header?

internal/provider/webhook_recipient_resource.go
```diff
@@ -372,16 +449,21 @@ func (r *webhookRecipientResource) Update(ctx context.Context, req resource.Upda
 	} else {
 		state.Secret = types.StringNull()
 	}
+
+	// to prevent confusing if/else blocks, set null by default and override it if we have that detail on the recipient
```
cewkrupa, line 453: Oooo this is a good idea, nice

client/recipient.go
brookesargent: I copied this from our external package struct but was wondering similarly if we wanted to switch it over. Personally for me, `Key` is clearer when speaking about HTTP headers because that's what things like Postman etc use but not a super strongly held opinion

cewkrupa: Ahh gotcha that's my bad then, matching the external struct is definitely what we want. I don't have a strong opinion on it and we can change it if needed... Names are hard

cewkrupa (approved):
While I can't speak to better ways to deal with the empty set of webhook headers, I think the rest of this looks excellent! Ty!

client/recipient.go
jharley: I think most of the omitempty shouldn't be there, tbh. They're not harming anything because of the way the client is used by the provider but they're probably not strictly "correct" either

client/recipient_test.go
```diff
@@ -130,6 +131,10 @@ func TestRecipientsCustomWebhook(t *testing.T) {
 			assert.Equal(t, tr.Details.WebhookSecret, r.Details.WebhookSecret)
 			assert.Equal(t, tr.Details.WebhookPayloads, r.Details.WebhookPayloads)
 			assert.Equal(t, tr.Details.WebhookPayloads.TemplateVariables, r.Details.WebhookPayloads.TemplateVariables)
+			require.Len(t, tr.Details.WebhookHeaders, 1)
+			require.Len(t, r.Details.WebhookHeaders, 1)
```
jharley, line 135: nit: this is maybe better written this way
```suggestion
			if assert.Len(t, r.Details.WebhookHeaders, 1) {
```

internal/provider/webhook_recipient_resource.go
```diff
@@ -34,6 +37,10 @@ var (
 
 	webhookTemplateTypes     = []string{"trigger", "exhaustion_time", "budget_rate"}
 	webhookTemplateNameRegex = regexp.MustCompile(`^[a-z](?:[a-zA-Z0-9]+$)?$`)
+
+	defaultHeaderContentType            = "Content-Type"
+	defaultHeaderUserAgent              = "User-Agent"
+	defaultHeaderXHoneycombWebhookToken = "X-Honeycomb-Webhook-Token"
```
jharley, line 43: if you pull these into a `[]string` (similar to to `webhookTemplateTypes` above) you can make use of [stringvalidator.NoneOfCaseInsensitive](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework-validators@v0.15.0/stringvalidator#NoneOfCaseInsensitive) and remove 276-286 in `ValidateConfig`

internal/provider/webhook_recipient_resource.go
```diff
@@ -145,6 +152,30 @@ func (*webhookRecipientResource) Schema(_ context.Context, _ resource.SchemaRequ
...
+							Description: "The name or key for the header",
+							Validators: []validator.String{
+								stringvalidator.LengthBetween(1, 64),
+							},
+						},
+						"value": schema.StringAttribute{
+							Description: "Value for the header",
+							Optional:    true,
+							Validators: []validator.String{
+								stringvalidator.LengthAtMost(256),
```
jharley, line 173: just matching the discussion above :)
```suggestion
								stringvalidator.LengthAtMost(512),
```

internal/provider/webhook_recipient_resource_test.go
jharley, line 1: with `value` being option in header blocks can you add a case where it's not-set?

internal/provider/webhook_recipient_resource.go
brookesargent: oooh nice! I wrongly assumed there wouldn't be a case insensitive built in

jharley (approved):
✨

PR #588, opened 2024-12-06
chore(ci): remove codecov

jharley:
We inherited codecov when we took this project over from the community, but haven't found it useful and decided to remove it from the project.

As followup once this is landed I'll remove the secret from the project's secrets.


cewkrupa (approved):
🧹 🧹 🧹

PR #622, opened 2025-02-21
feat(r/derived_column): expression syntax validation

### LINKED ISSUE #310: Validate derived column expressions during the planning stage
**Is your feature request related to a problem? Please describe.**

I'm sure this is not a trivial problem, but it comes up often enough in our company that I figured I'd mention it here.
When creating SLIs we don't get failures (e.g. syntax) until the change is being applied. Planning doesn't surface things like:

> Error: 400 Bad Request: invalid arguments for op IF, got 1, expected 2 or more

Since we use [Atlantis](https://www.runatlantis.io/) to apply changes via GitHub integration, this gets super weird for developers. They make changes and Atlantis shows them a "clean" plan, so they put it out for review. If the reviewer isn't eagle-eyed enough to catch the error, they approve the PR. The author then tells Atlantis to go ahead and apply, at which point it fails. 

Now the developer needs to figure out what's wrong (can often involve going to the Honeycomb UI to create a derived column with the expression from their Terraform file to debug), fix it, re-submit it for review/plan, and then re-attempt to apply the change.

This might be easier if Honeycomb allowed "inline" expressions in the query builder (allowing the developer to skip the step of creating a derived column and cleaning it up later), which is something we've been requesting for years. Even better would be to have that and to have the Terraform provider fail at plan time. But I'd settle for either soonish if I couldn't have both right away.

**Describe the solution you'd like**

For `terraform plan` to raise errors for invalid derived column expressions.

**Describe alternatives you've considered**

Only ever writing perfect derived column expressions. 😆 

**Additional context**

None.

jharley:
Adds syntax validation to `r/derived_column` during the resource's "validate" stage. 

- Closes #310


cewkrupa (approved):
Hooray! This is great

PR #625, opened 2025-02-28
fix(r/slo): force re-creation of SLO on SLI change

jharley:
Renaming a DC in use as an SLI is denied as it is in use. In an effort make it easier, we'll more or less match the UI behaviour and re-create an SLO if it's SLI changes.


jharley:
Closing this out in favour of using -- and documenting -- the `create_before_destroy` [lifecycle argument](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#create_before_destroy)

PR #635, opened 2025-03-13
feat(slo): add md slo support for slo data source

ronybzzz:
https://app.asana.com/1/72322038700732/project/1205961717360815/task/1205427815943458?focus=true

Adds Multi Dataset support for SLO data source 
Added unit tests to ensure SLO  read with `data honeycombio_slo` is still possible with optional dataset argument for MD SLOs

## Manual Verifications:
MD SLO:
Screenshot 2025-03-14 at 2 07 01 PM

Single SLO w/dataset argument:
Screenshot 2025-03-14 at 2 03 32 PM

Single SLO w/o dataset argument:
Screenshot 2025-03-14 at 2 02 15 PM


ronybzzz:
@jharley in the screenshot the dataset field returns a null value. should we return something to the dataset field when its omitted from the config? spoke with Madhu about it but we're not sure if anything might be relying on a value being returned? and if so, what would we make the value for MD SLOs, would it just be "__all__"?

internal/provider/slo_data_source_test.go
```diff
@@ -59,3 +60,86 @@ data "honeycombio_slo" "test" {
...
+func TestAcc_MDSLODataSource(t *testing.T) {
+	ctx := context.Background()
+	c := testAccClient(t)
+
+	if c.IsClassic(ctx) {
+		t.Skip("Classic does not support multi-dataset SLOs")
+	}
+
+	dataset_all := "__all__"
+	dataset1, err := c.Datasets.Create(ctx, &client.Dataset{
```
jharley, line 73: nit: could we prefix all of these test datasets, DCs, and SLOs with `test.`? There's a helper to do it (e.g. `test.RandomStringWithPrefix("test.", 10)`). It'll make automated cleanup far easier going forward

jharley (commented):
Great first pass at this! A few comments for clarity and tiding up.

> in the screenshot the dataset field returns a null value. should we return something to the dataset field when its omitted from the config?

I think that's to be expected -- you can even assert it in your MD SLO test. There's no way to not return an attribute in Terraform as far as I know and `null` feels more correct to me than `all`

MadhuVK (commented):
No additional comments from me, I think Jason's covered it :D

MadhuVK (approved):
Looks good from my side! Just one lingering reference to clean up I think based on prior feedback.

docs/data-sources/slos.md
```diff
@@ -1,7 +1,10 @@
 # Data Source: honeycombio_slos
-
+ for this data source
```
jharley, line 2:
```suggestion

```

jharley (approved):
Nice work! ✨

PR #638, opened 2025-03-18
chore(ci): don't highlight skipped tests

jharley:
Skipped tests are about to become more common as we have some features that only work in E&S vs Classic teams.

No need to highlight on the CI summary page.


robbkidd (approved):
Let's just skip it.

PR #641, opened 2025-03-19
feat(slo): add md slo support for resource slo

ronybzzz:
https://app.asana.com/0/1205961717360815/1207770326676046/f

Adds multi-dataset support for resource/slo. Deprecates the `dataset` argument and adds `datasets` to replace it. These arguments conflict with each other where at least one of these arguments are required however. 

Added unit tests, but you can also verify manually by creating a TF file that uses the datasets argument with at least 1 dataset and verify creating/updating/deleting works without needing to use `dataset` argument:
Screenshot 2025-03-19 at 4 13 26 PM

### DIFF
--- a/docs/resources/slo.md
+++ b/docs/resources/slo.md
@@ -21,7 +21,7 @@ resource "honeycombio_derived_column" "request_latency_sli" {
 resource "honeycombio_slo" "slo" {
   name              = "Latency SLO"
   description       = "example of an SLO"
-  dataset           = var.dataset
+  datasets          = [var.dataset]
   sli               = honeycombio_derived_column.request_latency_sli.alias
   target_percentage = 99.9
   time_period       = 30
@@ -37,7 +37,11 @@ The following arguments are supported:
 
 * `name` - (Required) The name of the SLO.
 * `description` - (Optional) A description of the SLO's intent and context.
-* `dataset` - (Required) The dataset this SLO is created in. Must be the same dataset as the SLI unless the SLI's dataset is `"__all__"`.
+* `dataset` - (Deprecated, conflicts with `datasets`) The dataset this SLO is created in. Must be the same dataset as the SLI unless the SLI's dataset is `"__all__"`.
+  * If `dataset` is not configured, `datasets` must be configured. 
+* `datasets` - (Required, conflicts with `dataset`) Array of datasets the SLO is evaluated on.
+  * If `datasets` is not configured, `dataset` must be configured.
+  * Must be >= 1 and <= 10. 
 * `sli` - (Required) The alias of the Derived Column that will be used as the SLI to indicate event success.
 The derived column used as the SLI must be in the same dataset as the SLO. Additionally,
 the column evaluation should consistently return nil, true, or false, as these are the only valid values for an SLI.
--- a/honeycombio/resource_slo.go
+++ b/honeycombio/resource_slo.go
@@ -39,9 +39,26 @@ func newSLO() *schema.Resource {
 			},
 			"dataset": {
 				Type:        schema.TypeString,
-				Required:    true,
+				Optional:    true,
 				ForceNew:    true,
 				Description: "The dataset this SLO is created in. Must be the same dataset as the SLI unless the SLI's dataset is `\"__all__\"`.",
+				DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool {
+					if oldValue == newValue {
+						return true
+					}
+					if newValue == "__all__" {
+						return true
+					}
+					return false
+				},
+			},
+			"datasets": {
+				Type:        schema.TypeSet,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Optional:    true,
+				Computed:    true,
+				ForceNew:    true,
+				Description: "The datasets the SLO is evaluated on.",
 			},
 			"sli": {
 				Type:     schema.TypeString,
@@ -62,6 +79,32 @@ the column evaluation should consistently return nil, true, or false, as these a
 				Description:  "The time period, in days, over which your SLO will be evaluated.",
 				ValidateFunc: validation.IntAtLeast(1),
 			},
+		},
+
+		CustomizeDiff: func(ctx context.Context, d *schema.ResourceDiff, meta interface{}) error {
+			dataset, datasetSet := d.GetOk("dataset")
+			datasets, datasetsSet := d.GetOk("datasets")
+
+			slugs := datasets.(*schema.Set).List()
+
+			// If dataset is explicitly set (not the default) and more than 1 datasets is set, return an error
+			if datasetsSet && len(slugs) > 1 && datasetSet && dataset != "__all__" {
+				return errors.New("if more than 1 'datasets' is set, 'dataset' must be left as default ('__all__') or unset")
+			}
+
+			// not sure how it would happen, but adding validation in case
+			if datasetSet && len(slugs) == 1 && dataset != "__all__" && slugs[0] != dataset {
+				return errors.New("'dataset' must be the same as the only element of 'datasets'")
+			}
+
+			// Validate datasets length if it's provided or if dataset was set to "__all__"
+			if datasetsSet || dataset == "__all__" {
+				if len(slugs) < 1 || len(slugs) > 10 {
+					return errors.New("'datasets' must contain between 1 and 10 datasets")
+				}
+			}
+
+			return nil
 		},
 	}
 }
@@ -85,7 +128,8 @@ func resourceSLOCreate(ctx context.Context, d *schema.ResourceData, meta interfa
 		return diagFromErr(err)
 	}
 
-	dataset := d.Get("dataset").(string)
+	dataset := getDataset(d)
+
 	s, err := client.SLOs.Create(ctx, dataset, expandSLO(d))
 	if err != nil {
 		return diag.FromErr(err)
@@ -101,7 +145,7 @@ func resourceSLORead(ctx context.Context, d *schema.ResourceData, meta interface
 		return diagFromErr(err)
 	}
 
-	dataset := d.Get("dataset").(string)
+	dataset := getDataset(d)
 
 	var detailedErr honeycombio.DetailedError
 	s, err := client.SLOs.Get(ctx, dataset, d.Id())
@@ -122,6 +166,7 @@ func resourceSLORead(ctx context.Context, d *schema.ResourceData, meta interface
 	d.Set("sli", s.SLI.Alias)
 	d.Set("target_percentage", helper.PPMToFloat(s.TargetPerMillion))
 	d.Set("time_period", s.TimePeriodDays)
+	d.Set("datasets", s.DatasetSlugs)
 
 	return nil
 }
@@ -132,7 +177,8 @@ func resourceSLOUpdate(ctx context.Context, d *schema.ResourceData, meta interfa
 		return diagFromErr(err)
 	}
 
-	dataset := d.Get("dataset").(string)
+	dataset := getDataset(d)
+
 	s, err := client.SLOs.Update(ctx, dataset, expandSLO(d))
 	if err != nil {
 		return diag.FromErr(err)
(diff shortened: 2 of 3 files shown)


honeycombio/resource_slo.go
```diff
@@ -158,12 +204,28 @@ func resourceSLODelete(ctx context.Context, d *schema.ResourceData, meta interfa
 }
 
 func expandSLO(d *schema.ResourceData) *honeycombio.SLO {
+	var datasets []string
+	if v, ok := d.GetOk("datasets"); ok {
+		for _, v := range v.(*schema.Set).List() {
+			datasets = append(datasets, v.(string))
```
ronybzzz, line 210: @jharley running into an issue after trying both `schema.TypeList` and `schema.TypeSet` for the datasets type. it seems like the datasets get created fine with the order the datasets were defined in the config, but if i run terraform apply again, it seems like the way its reading the SLO changes the order of the dataset array and causing it to look like an update. Any thoughts on solving this? i think this is whats causing the test the fail since the ordering gets mixed up. 
[
Screenshot 2025-03-19 at 4 43 27 PM
Screenshot 2025-03-19 at 4 43 37 PM

honeycombio/resource_slo.go
```diff
@@ -39,9 +39,26 @@ func newSLO() *schema.Resource {
...
 				Description: "The dataset this SLO is created in. Must be the same dataset as the SLI unless the SLI's dataset is `\"__all__\"`.",
+				DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool {
+					if oldValue == newValue {
+						return true
+					}
+					if newValue == "__all__" {
+						return true
+					}
+					return false
+				},
```
jharley, line 53: Given the way this has come together, I'm not sure this is needed 🤔 Both are marked 'ForceNew' so removing or updating one for the other should cause re-creation anyway?

Or is the intention to *not* force re-creation and update all SLOs to have `__all__` in state regardless of the written value? That's not what is currently happening so checking on it. If that is the plan there may be some issues reconciling what is in state with what is written in the config.

docs/resources/slo.md
```diff
@@ -37,7 +37,11 @@ The following arguments are supported:
 
 * `name` - (Required) The name of the SLO.
 * `description` - (Optional) A description of the SLO's intent and context.
-* `dataset` - (Required) The dataset this SLO is created in. Must be the same dataset as the SLI unless the SLI's dataset is `"__all__"`.
+* `dataset` - (Deprecated, conflicts with `datasets`) The dataset this SLO is created in. Must be the same dataset as the SLI unless the SLI's dataset is `"__all__"`.
```
jharley, line 40: I'd rather we _not_ mark it deprecated right away: a lot of people use this resource and they'll be getting warnings until they re-create all of their SLOs

we should instead make a note that it'll be deprecated in a future release

docs/resources/slo.md
```diff
@@ -37,7 +37,11 @@ The following arguments are supported:
...
+  * If `dataset` is not configured, `datasets` must be configured. 
+* `datasets` - (Required, conflicts with `dataset`) Array of datasets the SLO is evaluated on.
```
jharley, line 42: not strictly required: either this one or the other is required

honeycombio/resource_slo_test.go
```diff
@@ -88,7 +136,7 @@ func testAccCheckSLOExists(t *testing.T, name string, slo *honeycombio.SLO) reso
 		}
 
 		client := testAccClient(t)
-		rslo, err := client.SLOs.Get(context.Background(), resourceState.Primary.Attributes["dataset"], resourceState.Primary.ID)
+		rslo, err := client.SLOs.Get(context.Background(), "__all__", resourceState.Primary.ID)
```
jharley, line 139: nit
```suggestion
		rslo, err := client.SLOs.Get(context.Background(), honeycombio.EnvironmentWideSlug, resourceState.Primary.ID)
```

honeycombio/resource_slo_test.go
```diff
@@ -119,3 +167,55 @@ func sloAccTestSetup(t *testing.T) (string, string) {
...
+			Settings: honeycombio.DatasetSettings{
+				DeleteProtected: helper.ToPtr(false),
+			},
+		})
+		err = c.Datasets.Delete(ctx, dataset2.Slug)
+		require.NoError(t, err)
+	})
+
+	sli, err := c.DerivedColumns.Create(ctx, honeycombio.EnvironmentWideSlug, &honeycombio.DerivedColumn{
+		Alias:       "sli." + acctest.RandString(8),
```
jharley, line 210:
```suggestion
		Alias:       test.RandomStringWithPrefix("test.", 10),
```

docs/resources/slo.md
```diff
@@ -37,7 +37,11 @@ The following arguments are supported:
...
+  * If `datasets` is not configured, `dataset` must be configured.
+  * Must be >= 1 and <= 10.
```
jharley, line 44: nit: this nested-bullet formatting is not used elsewhere in the docs, so I'd rather we be consistent. Take a look at the trigger docs for an example listing conflicts and details like "up to 10"

honeycombio/resource_slo.go
MadhuVK: My best guess how to resolve:

resourceSLORead populates datasets via a slice. I think converting into a Set might be what we need.

I've minimially used sets though, so even if that works, I don't know if it's good practice to use Sets. Perhaps deterministic sort-ordering is better?

jharley: It's currently a Set and appears to be working?

honeycombio/resource_slo.go
MadhuVK: I think this is on me for not explaining my intent correctly.

Main thing I'm interested in: If the practitioner moves from the deprecated field (`Dataset`) to the new field (`Datasets`), is there a way they can do so without forcing a new resource?
* The deprecation warnings can be pretty annoying.
* And if there's a clean way to migrate forward and illustrate "these states are conceptually the same," that would be ideal.

honeycombio/resource_slo.go
MadhuVK: Specifically this line is what I was thinking: https://github.com/honeycombio/terraform-provider-honeycombio/pull/641/files#diff-a779b2f9a4fea5862bd3e702a800f92d587749899277e80892c383fd6183f62aR169

If you notice in Rony's first screenshot, the apply output seems to be rendering `datasets` as a list, rather than a Set as on the schema.

ronybzzz: hmm i mightve included the wrong screenshot. i think it might be working? its only when i see the printed result using the below, that i see the order get sorted. but if i hit apply multiple times, no change is detected which is the correct behavior
```
output "result" {
value = resource.honeycombio_slo.md_slo
}
```

jharley: It's working for me in the current form with `TypeSet`

honeycombio/resource_slo.go
jharley: I think that's a great intention, and worth testing

jharley: coming back to this: can the datasets an SLO is looking at be updated in place? The current schema prevents that? (`ForceNew` is set)

honeycombio/resource_slo.go
```diff
@@ -39,9 +39,31 @@ func newSLO() *schema.Resource {
...
+				DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool {
+					if oldValue == newValue {
+						return true
+					}
+					// if the config moves away from deprecated dataset, nothing should change
+					if newValue == "" || newValue == "__all__" {
+						return true
+					}
+					return false
+				},
```
jharley, line 54: If we're going to ship this functionality, I'd like to see a test for it.

We're also not currently setting an MDSLOs dataset to `__all__` (you're asserting the property is empty below.

honeycombio/resource_slo.go
```diff
@@ -39,9 +39,31 @@ func newSLO() *schema.Resource {
...
+			"datasets": {
+				Type:         schema.TypeSet,
+				Elem:         &schema.Schema{Type: schema.TypeString},
+				Optional:     true,
+				Computed:     true,
+				ForceNew:     true,
+				Description:  "The datasets the SLO is evaluated on.",
+				ExactlyOneOf: []string{"dataset", "datasets"},
+				MaxItems:     10,
+				MinItems:     1,
```
jharley, line 66: when `Optional` is true, `MinItems` is ignored
```suggestion
```

jharley (commented):
Lookin' pretty great! I'm mostly curious if we're going to pursue the "you can switch from dataset to datasets" in place approach and if so I'd like some tests in place to ensure it works (and continues to)

honeycombio/resource_slo.go
ronybzzz: for datasets, i think no? but for dataset, if it changes to "__all__" then yes, it just becomes null in the state. @MadhuVK correct me if im wrong

honeycombio/resource_slo.go
ronybzzz: hmm when i remove this line, it allows me to pass an empty array to datasets which we dont want. this seemed to have fixed that validation

jharley: Huh.. I'm pulling that from `MinItems` [documentation](https://github.com/hashicorp/terraform-plugin-sdk/blob/b8445ee6b61bd574c02047f3686c74e4d9b4348a/helper/schema/schema.go#L252-L257)

Let's put it back in, then, with my apologies :)

honeycombio/resource_slo.go
jharley: Great! We need an acceptance test written that does what you just did manually then, please. This functionality should be written in the docs for the resource as well as it's otherwise just "hidden magic".

I'm also asking if it's possible to go from `datasets = ["rony-test", "jason-test"]` from just `rony-test` (which, even if we support API side the `ForceNew` will do a re-creation). Is the goal purely a migration one for the future when `dataset` is deprecated? Cool if so, but we need to cover it in documentation (user facing and in the code) and test coverage :)

ronybzzz: oh i dont think we can update datasets in place like that, i think its just for single dataset slo being able to use either `dataset` or a single array `datasets`, @MadhuVK correct me if im wrong!

docs/resources/slo.md
```diff
@@ -37,13 +64,16 @@ The following arguments are supported:
 
 * `name` - (Required) The name of the SLO.
 * `description` - (Optional) A description of the SLO's intent and context.
-* `dataset` - (Required) The dataset this SLO is created in. Must be the same dataset as the SLI unless the SLI's dataset is `"__all__"`.
+* `dataset` - (Optional) The dataset this SLO is created in. Must be the same dataset as the SLI unless the SLI's dataset is `"__all__"`. Conflicts with `datasets`.
```
jharley, line 67:
```suggestion
* `dataset` - (Optional) The dataset this SLO is created in. Must be the same dataset as the SLI unless the SLI's dataset is `"__all__"`. Conflicts with `datasets`. Will be deprecated in a future release of the provider.
```

jharley (approved):
This is looking solid! Thank you

honeycombio/resource_slo.go
```diff
@@ -39,9 +39,31 @@ func newSLO() *schema.Resource {
 			},
 			"dataset": {
 				Type:        schema.TypeString,
-				Required:    true,
+				Optional:    true,
 				ForceNew:    true,
 				Description: "The dataset this SLO is created in. Must be the same dataset as the SLI unless the SLI's dataset is `\"__all__\"`.",
+				DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool {
```
jharley, line 45: Thought about this a bit, and I like the pattern especially combined with the `getDataset` helper function as a way to drop `dataset` as a requirement for things that have environment-wide support -- and by extension remove the need to have users of the provider do `__all__` 😅 

Derived Columns most notably would benefit from `dataset` becoming optional, but I'm sure with a quick-ish audit I can find others.

To be clear: not for you to resolve here, but just saying I appreciate the thoughtful attention to the developer experience ✨

PR #642, opened 2025-03-24
feat(d/query_specification): 'order' validation support

jharley:
Validation of `order` blocks was missing from `d/query_specification`, allowing the potential of apply-time errors that could have been caught during validation/planning.


internal/provider/query_specification_data_source_test.go
```diff
@@ -210,6 +211,7 @@ data "honeycombio_query_specification" "test" {
     column = "we-should-not-specify-a-column-with-COUNT"
   }
 }`,
+		PlanOnly:    true,
```
jharley, line 214: All of these `PlanOnly: true` additions are just to speed things up here a bit: no external resources need to be created.

cewkrupa (approved):

PR #644, opened 2025-03-24
feat(r/burn_alert): add multi-dataset support

ronybzzz:

- Closes https://app.asana.com/0/1205961717360815/1207770326676048/f

- adds multi dataset support for r/burn_alerts by making the `dataset` argument optional.

- add unit tests
- you can check out the **slo_with_burn_alert** example and omit the `dataset` arg. 

#### Screenshots of terraform plan when omitting `dataset` for burn alert resource:

Screenshot 2025-03-24 at 2 19 23 PM
Screenshot 2025-03-24 at 2 20 01 PM

### DIFF
--- a/docs/resources/burn_alert.md
+++ b/docs/resources/burn_alert.md
@@ -1,13 +1,14 @@
 # Resource: honeycombio_burn_alert
 
-Creates a burn alert. 
+Creates a burn alert.
 
-For more information about burn alerts, 
+For more information about burn alerts,
 check out [Define Burn Alerts](https://docs.honeycomb.io/working-with-your-data/slos/burn-alerts).
 
 ## Example Usage
 
 ### Basic Example - Exhaustion Time Burn Alert
+
 ```hcl
 variable "dataset" {
   type = string
@@ -39,6 +40,7 @@ resource "honeycombio_burn_alert" "example_alert" {
 ```
 
 ### Basic Example - Budget Rate Burn Alert
+
 ```hcl
 variable "dataset" {
   type = string
@@ -66,6 +68,7 @@ resource "honeycombio_burn_alert" "example_alert" {
 ```
 
 ### Example - Exhaustion Time Burn Alert with PagerDuty Recipient and Severity
+
 ```hcl
 variable "dataset" {
   type = string
@@ -101,6 +104,7 @@ resource "honeycombio_burn_alert" "example_alert" {
 ```
 
 ### Example - Exhaustion Time Burn Alert with Webhook Recipient and Notification Variable
+
 ```hcl
 variable "dataset" {
   type = string
@@ -110,7 +114,7 @@ variable "slo_id" {
   type = string
 }
 
-data "honeycombio_recipient" "custom_webhook" { 
+data "honeycombio_recipient" "custom_webhook" {
     type = "webhook"
 
     detail_filter {
@@ -119,7 +123,7 @@ data "honeycombio_recipient" "custom_webhook" {
     }
 }
 
-resource "honeycombio_burn_alert" "example_alert" { 
+resource "honeycombio_burn_alert" "example_alert" {
     exhaustion_minutes = 60
     description        = "Burn alert description"
     dataset            = var.dataset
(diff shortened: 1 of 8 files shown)


internal/provider/burn_alert_resource.go
```diff
@@ -97,11 +97,12 @@ func (*burnAlertResource) Schema(_ context.Context, _ resource.SchemaRequest, re
 				},
```
ronybzzz, line 97: is there any special handling for Import? or is it just left as is similar to slo? not really sure how import works..

internal/provider/burn_alert_resource.go
```diff
@@ -450,3 +463,10 @@ func (r *burnAlertResource) Delete(ctx context.Context, req resource.DeleteReque
 		}
 	}
 }
+
+func getDatasetString(dataset types.String) types.String {
```
jharley, line 467: this should be moved to `internal/helper`

internal/provider/burn_alert_resource_test.go
```diff
@@ -931,6 +972,36 @@ resource "honeycombio_burn_alert" "test" {
 }`, exhaustionMinutes, dataset, sloID, pdseverity, testBADescription, tmplBody)
 }
 
+func testAccConfigBurnAlertDefault_MD(exhaustionMinutes int, sloID, pdseverity string) string {
+	tmplBody := `<<EOT
```
jharley, line 976: I don't think this is used (or needed)?

internal/provider/burn_alert_resource.go
```diff
@@ -97,11 +97,12 @@ func (*burnAlertResource) Schema(_ context.Context, _ resource.SchemaRequest, re
 				},
 			},
 			"dataset": schema.StringAttribute{
-				Required:    true,
+				Optional:    true,
 				Description: "The dataset this Burn Alert is associated with.",
 				PlanModifiers: []planmodifier.String{
 					stringplanmodifier.RequiresReplace(),
 				},
+				DeprecationMessage: "No longer required.",
```
jharley, line 105: We haven't yet deprecated it
```suggestion
```

internal/provider/burn_alert_resource.go
```diff
@@ -97,11 +97,12 @@ func (*burnAlertResource) Schema(_ context.Context, _ resource.SchemaRequest, re
 				},
 			},
 			"dataset": schema.StringAttribute{
-				Required:    true,
+				Optional:    true,
 				Description: "The dataset this Burn Alert is associated with.",
 				PlanModifiers: []planmodifier.String{
```
jharley, line 102: Should we be doing a similar treatment here where we suppress the diff if `dataset` is unset?

internal/provider/burn_alert_resource.go
jharley: This question actually made me wonder if MDSLO imports are working: I think they may not be as they're still requiring `datasetSlug/ID` and we need to handle both forms now (bare `ID` and `dataset/ID`).

I think of imports as skipping Create by adding the ID into state and then having a read called to reconcile against the configuration.

internal/provider/burn_alert_resource.go
ronybzzz: do you know how i can add the diff suppress function here when using the framework rather than the sdk?

jharley: In the Plugin Framework they're called [PlanModifiers](https://developer.hashicorp.com/terraform/plugin/framework/resources/plan-modification). A migration guide comparing them to the [CustomizeDiff](https://developer.hashicorp.com/terraform/plugin/framework/migrating/resources/plan-modification#framework) approach in SDKv2

ronybzzz:
synced with @MadhuVK and settled on leaving the state with the old value if going from "test-dataset" to omitting, the plan shows no changes so its as close to the behavior as we can get.

honeycombio/resource_slo.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 
+	"github.com/honeycombio/terraform-provider-honeycombio/client"
```
jharley, line 12: this package is already imported as `honeycombio`: a bit odd but is standard in the SDKv2 code so it'd be good to stick with it

internal/provider/burn_alert_resource.go
```diff
@@ -226,6 +243,14 @@ func (r *burnAlertResource) ImportState(ctx context.Context, req resource.Import
 		return
 	}
 
+	if dataset == client.EnvironmentWideSlug {
+		resp.Diagnostics.Append(resp.State.Set(ctx, &models.BurnAlertResourceModel{
+			ID:         types.StringValue(id),
```
jharley, line 248: I'm curious if we should be explicitly setting dataset to Null or empty string here for clarity? 🤔

docs/resources/burn_alert.md
```diff
@@ -192,6 +192,8 @@ In addition to all arguments above, the following attributes are exported:
 
 Burn alerts can be imported using a combination of the dataset name and their ID, e.g.
 
+For multi-dataset burn alerts, replace dataset name with `__all__`.
```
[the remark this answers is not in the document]
ronybzzz: good point!

internal/provider/burn_alert_resource.go
ronybzzz: i think thats a good idea

docs/resources/burn_alert.md
```diff
@@ -192,6 +192,8 @@ In addition to all arguments above, the following attributes are exported:
...
+
 ```
 $ terraform import honeycombio_burn_alert.my_alert my-dataset/bj9BwOb1uKz
 ```
```
jharley, line 199:
```suggestion

Burn alerts can be imported using by using their ID combined with their dataset.
If the burn alert is a multi-dataset (MD) burn alert, the dataset is not provided.

### Burn Alert

```
$ terraform import honeycombio_burn_alert.my_alert my-dataset/bj9BwOb1uKz
```

### Multi-dataset Burn Alert

```
$ terraform import honeycombio_burn_alert.my_alert bc9XwOb2yJu
```

```

docs/resources/slo.md
```diff
@@ -84,6 +84,8 @@ In addition to all arguments above, the following attributes are exported:
 
 SLOs can be imported using a combination of the dataset name and their ID, e.g.
 
+For multi-dataset SLOs, replace dataset name with `__all__`.
```
jharley, line 87: This should be updated to something similar to the above

honeycombio/resource_slo.go
```diff
@@ -90,12 +90,17 @@ the column evaluation should consistently return nil, true, or false, as these a
 
 func resourceSLOImport(ctx context.Context, d *schema.ResourceData, i interface{}) ([]*schema.ResourceData, error) {
 	// import ID is of the format <dataset>/<SLO ID>
+	// To import MD SLOs, just pass in the <SLO ID>
```
jharley, line 93:
```suggestion
```

honeycombio/resource_slo.go
```diff
@@ -90,12 +90,17 @@ the column evaluation should consistently return nil, true, or false, as these a
...
 	dataset, id, found := strings.Cut(d.Id(), "/")
+
+	// if separator not found, we will assume its the bare id
```
jharley, line 96:
```suggestion
	// if dataset separator not found, we will assume its the bare id
```

honeycombio/resource_slo_test.go
```diff
@@ -132,9 +132,21 @@ func TestHoneycombSLO_MD(t *testing.T) {
 					resource.TestCheckResourceAttr("honeycombio_slo.md_test", "time_period", "30"),
 				),
 			},
+			// tests imports
+			{
+				ResourceName:      "honeycombio_slo.md_test",
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateIdFunc: func(s *terraform.State) (string, error) {
```
jharley, line 140: as far as I know we haven't needed an `ImportStateIdFunc` previously. Do we need this one? If we do, a comment as to why would be good

internal/provider/burn_alert_resource.go
```diff
@@ -217,12 +233,18 @@ func (r *burnAlertResource) ValidateConfig(ctx context.Context, req resource.Val
 
 func (r *burnAlertResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
 	// import ID is of the format <dataset>/<BurnAlert ID>
+	// To import MD burn alerts, just pass in <BurnAlert ID>
```
jharley, line 236:
```suggestion
```

internal/provider/burn_alert_resource.go
```diff
@@ -217,12 +233,18 @@ func (r *burnAlertResource) ValidateConfig(ctx context.Context, req resource.Val
...
 	dataset, id, found := strings.Cut(req.ID, "/")
+
+	// if separator not found, we will assume its the bare id
```
jharley, line 239:
```suggestion
	// if dataset separator not found, we will assume its the bare id
```

internal/provider/burn_alert_resource.go
```diff
@@ -217,12 +233,18 @@ func (r *burnAlertResource) ValidateConfig(ctx context.Context, req resource.Val
...
+	// if thats the case, we need to reassign values since strings.Cut would return (id, "", false)
 	if !found {
```
jharley, line 241: nit: little less duplication written this way

```go
	dsValue := types.StringNull()
	if !found {
		id = dataset
	} else {
		dsValue = types.StringValue(dataset)
	}

	resp.Diagnostics.Append(resp.State.Set(ctx, &models.BurnAlertResourceModel{
		ID:         types.StringValue(id),
		Dataset:    dsValue,
		Recipients: types.SetUnknown(types.ObjectType{AttrTypes: models.NotificationRecipientAttrType}),
	})...)
```

internal/provider/burn_alert_resource_test.go
```diff
@@ -324,33 +324,6 @@ func TestAcc_BurnAlertResourceUpgradeFromVersion015(t *testing.T) {
 	})
 }
 
-func TestAcc_BurnAlertResource_Import_validateImportID(t *testing.T) {
```
jharley, line 327: I think it might be worth leaving this test in place to make sure we don't break `dataset/ID` imports

internal/provider/burn_alert_resource_test.go
```diff
@@ -849,11 +822,66 @@ func testAccEnsureBurnAlertDestroyed(t *testing.T) resource.TestCheckFunc {
...
+			},
+			// Update - PD Severity from info -> critical (the default)
+			{
+				Config: testAccConfigBurnAlertDefault_MD(exhaustionMinutes, sloID, "critical"),
+				Check:  testAccEnsureSuccessExhaustionTimeAlert(t, burnAlert, exhaustionMinutes, "critical", sloID),
+			},
+			// Import
+			{
+				ResourceName:            "honeycombio_burn_alert.test",
+				ImportStateIdPrefix:     "",
```
jharley, line 861:
```suggestion
```

jharley (commented):
I think you're nearly there! 🥳 

Last critical thing is to clean up the docs, but a few suggestions and a question

internal/provider/burn_alert_resource_test.go
ronybzzz: hmmm so before it would return this error if <dataset>/<id> wasnt able to be parse. Now that we are assuming that if just <id> is provided its a md resource, would we just be trying to expect a different error such as: "Error: Cannot import non-existent remote object"

jharley: Oh! I'm sorry: I read the test wrong and it's testing the _error_ state. My fault: thank you for clarifying

internal/provider/burn_alert_resource.go
```diff
@@ -97,10 +98,25 @@ func (*burnAlertResource) Schema(_ context.Context, _ resource.SchemaRequest, re
 				},
 			},
 			"dataset": schema.StringAttribute{
-				Required:    true,
+				Optional:    true,
+				Computed:    true,
 				Description: "The dataset this Burn Alert is associated with.",
 				PlanModifiers: []planmodifier.String{
-					stringplanmodifier.RequiresReplace(),
+					modifiers.DatasetDeprecation(),
```
jharley, line 105: final thing: can we get some test coverage for the cases where there planmodifiers run?

jharley (commented):
This is looking excellent! Thank you for the effort.

A final bit of test coverage needed then we're good to go!

internal/provider/burn_alert_resource_test.go
```diff
@@ -66,6 +66,13 @@ func TestAcc_BurnAlertResource_defaultBasic(t *testing.T) {
 				Config: testAccConfigBurnAlertBudgetRate_basic(budgetRateWindowMinutes, budgetRateDecreasePercent, dataset, sloID, "info"),
 				Check:  testAccEnsureSuccessBudgetRateAlert(t, burnAlert, budgetRateWindowMinutes, budgetRateDecreasePercent, "info", sloID),
 			},
+
+			// update the config to remove dataset and ensure nothing changes
+			{
+				Config:             testAccConfigBurnAlertBudgetRate_basic_dataset_deprecation(budgetRateWindowMinutes, budgetRateDecreasePercent, sloID, "info"),
+				PlanOnly:           true,
+				ExpectNonEmptyPlan: false,
```
jharley, line 74: I think this passes fine without it?
```suggestion
```

internal/provider/burn_alert_resource.go
```diff
@@ -97,10 +98,25 @@ func (*burnAlertResource) Schema(_ context.Context, _ resource.SchemaRequest, re
...
+					stringplanmodifier.RequiresReplaceIf(
+						func(ctx context.Context, sr planmodifier.StringRequest, rrifr *stringplanmodifier.RequiresReplaceIfFuncResponse) {
+							// If the dataset is empty, we don't want to require a replace
```
jharley, line 108: running the test suite with coverage shows that this `RequiresReplaceIf` function isn't being executed at all 🤔 

Are we sure we need the function, or is the test not sufficient?

internal/provider/burn_alert_resource_test.go
```diff
@@ -878,6 +921,43 @@ func burnAlertAccTestSetup(t *testing.T) (string, string) {
...
+	ctx := context.Background()
+	c := testAccClient(t)
+
+	dataset, err := c.Datasets.Create(ctx, &client.Dataset{
+		Name: acctest.RandomWithPrefix("test"),
+	})
+	require.NoError(t, err)
+
+	t.Cleanup(func() {
+		c.Datasets.Delete(ctx, dataset.Slug)
```
jharley, line 936: I believe this delete is failing as deletion protection isn't disabled after creation above

internal/provider/burn_alert_resource_test.go
```diff
@@ -878,6 +921,43 @@ func burnAlertAccTestSetup(t *testing.T) (string, string) {
...
 }
 
+func getNewDatasetAndSLO(t *testing.T) (string, string) {
+	t.Helper()
+
+	ctx := context.Background()
+	c := testAccClient(t)
+
+	dataset, err := c.Datasets.Create(ctx, &client.Dataset{
+		Name: acctest.RandomWithPrefix("test"),
```
jharley, line 931: I'd rather see this version as the sweepers are cleaning things up that start with `test.` (this variation is making things like `test-3280250866423422428`)
```suggestion
		Name: test.RandomStringWithPrefix("test.", 10),
```

jharley (approved):
Thanks for working through this, @ronybzzz! 🚀

PR #648, opened 2025-03-31
chore(r/webhook_recipient): increase header value char limit

kattelles:

Increase the webhook header char limit to match the updated upstream API limit.


jharley (approved):
Thanks for keeping these in sync, @kattelles!

PR #651, opened 2025-04-03
fix(client): query result now returns empty series upstream

jharley:
Client was marshalling these to nil, but upstream is now returning empty arrays.

Upstream change is within boundary of documentation, just client test needs to be updated.


jharley:
`main` has been building fine without this.. so I'm not sure what the flake was. Will close this for now and keep an eye on it 🤔

PR #652, opened 2025-04-03
fix(build): remove incorrect 'toolchain' directive

jharley:
dependabot incorrectly added a Go 1.24 `toolchain` directive in #650

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,8 +2,6 @@ module github.com/honeycombio/terraform-provider-honeycombio
 
 go 1.23.8
 
-toolchain go1.24.1
-
 require (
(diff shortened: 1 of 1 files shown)


cewkrupa (approved):
Ugh. Is this something we can configure with dependabot maybe?

jharley:
> Ugh. Is this something we can configure with dependabot maybe?

Gonna look into that next

jharley:
> Ugh. Is this something we can configure with dependabot maybe?

It's a known issue it seems: https://github.com/dependabot/dependabot-core/issues/11933

PR #654, opened 2025-04-04
feat: remove need for magic environment-wide string

jharley:
Previously, creating "environment wide" resources required the use of the ✨ magic ✨  `dataset = "__all__"`.

With this change, resources that have Environment-wide support will move to the "dataset" argument becoming optional with the assumption that if it isn't set the intention is to operate at the Environment level.

### DIFF
--- a/docs/resources/derived_column.md
+++ b/docs/resources/derived_column.md
@@ -24,7 +24,7 @@ The following arguments are supported:
 
 * `dataset` - (Required) The dataset this derived column is added to. If not set, an Environment-wide derived column will be created.
 * `alias` - (Required) The name of the derived column. Must be unique per dataset.
-* `expression` - (Required) The formula of the derived column. See [Derived Column Syntax](https://docs.honeycomb.io/working-with-your-data/customizing-your-query/derived-columns/#derived-column-syntax).
+* `expression` - (Required) The formula of the derived column. See [Derived Column Syntax](https://docs.honeycomb.io/reference/derived-column-formula/syntax/).
 * `description` - (Optional) A description that is shown in the UI.
 
 ## Attribute Reference
--- a/docs/resources/derived_column.md
+++ b/docs/resources/derived_column.md
@@ -22,9 +22,9 @@ resource "honeycombio_derived_column" "duration_ms_log" {
 
 The following arguments are supported:
 
-* `dataset` - (Required) The dataset this derived column is added to. Use `__all__` for Environment-wide derived columns.
+* `dataset` - (Required) The dataset this derived column is added to. If not set, an Environment-wide derived column will be created.
 * `alias` - (Required) The name of the derived column. Must be unique per dataset.
-* `expression` - (Required) The function of the derived column. See [Derived Column Syntax](https://docs.honeycomb.io/working-with-your-data/customizing-your-query/derived-columns/#derived-column-syntax).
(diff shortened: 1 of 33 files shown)


honeycombio/resource_derived_column.go
```diff
@@ -30,11 +31,13 @@ func newDerivedColumn() *schema.Resource {
...
 				ForceNew:     true,
+				Description:  "The alias of the derived column. Must be unique within the dataset or environment.",
 				ValidateFunc: validation.StringLenBetween(1, 255),
 			},
 			"expression": {
-				Type:     schema.TypeString,
-				Required: true,
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The formula of the derived column. See [Derived Column Syntax](https://docs.honeycomb.io/working-with-your-data/customizing-your-query/derived-columns/#derived-column-syntax).",
```
cewkrupa, line 40:
```suggestion
				Description: "The formula of the derived column. See [Derived Column Syntax](https://docs.honeycomb.io/reference/derived-column-formula/syntax/).",
```

honeycombio/resource_marker_test.go
```diff
@@ -31,35 +38,54 @@ func TestAccHoneycombioMarker_basic(t *testing.T) {
...
 		}
 
 		c := testAccClient(t)
-		m, err := c.Markers.Get(context.Background(), resourceState.Primary.Attributes["dataset"], resourceState.Primary.ID)
+		_, err := c.Markers.Get(context.Background(), dataset, resourceState.Primary.ID)
 		if err != nil {
 			return fmt.Errorf("could not retrieve marker: %w", err)
 		}
 
-		assert.Equal(t, "Hello world!", m.Message)
```
cewkrupa, line 59: why'd these bits get removed?

cewkrupa (approved):
Wow, this was a lot! I _think_ this makes sense and looks good. Few links to update but nothing jumping out at me as a problem!

honeycombio/resource_marker_test.go
jharley: Ah, I should have commented!

They're being asserted above using the native functionality already (so we're doing it twice): this function should just be asserting that the marker exists remotely

cewkrupa: Got it - thank you!

docs/resources/derived_column.md
```diff
@@ -22,9 +22,9 @@ resource "honeycombio_derived_column" "duration_ms_log" {
 
 The following arguments are supported:
 
-* `dataset` - (Required) The dataset this derived column is added to. Use `__all__` for Environment-wide derived columns.
+* `dataset` - (Required) The dataset this derived column is added to. If not set, an Environment-wide derived column will be created.
 * `alias` - (Required) The name of the derived column. Must be unique per dataset.
-* `expression` - (Required) The function of the derived column. See [Derived Column Syntax](https://docs.honeycomb.io/working-with-your-data/customizing-your-query/derived-columns/#derived-column-syntax).
+* `expression` - (Required) The formula of the derived column. See [Derived Column Syntax](https://docs.honeycomb.io/working-with-your-data/customizing-your-query/derived-columns/#derived-column-syntax).
```
jharley, line 27:
```suggestion
* `expression` - (Required) The formula of the derived column. See [Derived Column Syntax](https://docs.honeycomb.io/reference/derived-column-formula/syntax/).
```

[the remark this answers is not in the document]
jharley: Ah, thank you! Updated

PR #659, opened 2025-04-10
chore(r/board): migrate to Plugin Framework

### LINKED ISSUE #399: r/honeycombio_board: removing "graph_settings" from query does not set all values to false
Due to what I believe is a limitation of the Plugin SDKv2, when removing `graph_settings` from a board `query` the values will not be all set to `false` but instead left as they were previously in state.

Bug discovered while investigating another bug related to `graph_settings` (see: #400).

Likely fix is to reimplement the `honeycombio_board` resource in the new Plugin Framework (as has been done for the trigger and burn alert resource).

**Steps to reproduce**

Add a query to a board with `graph_settings` set, apply it, then after creation remove the query's `graph_settings` block. When you run plan/apply again there will be no changes to apply but `terraform show` retains the old values.

**Additional context**

Suggested workaround is to either leave `graph_settings {}` in the `query` or simply leave one or more settings as `false`. e.g.

```hcl
resource "honeycombio_board" "myboard" {
[...]
  query {
    query_id = honeycombio_query.myquery.id
  
    graph_settings {
      utc_xaxis = false
    }
  }
}
```

jharley:
Migrates `r/board` from the Plugin SDKv2 to the Plugin Framework.

In the name of a more "lift and shift" migration, this leaves the deprecated fields in place but does manage to close a long-open bug related to `graph_settings`.

- Closes #399 

---
  - https://app.asana.com/0/0/1209849296124978


internal/provider/board_resource_test.go
```diff
@@ -152,8 +153,6 @@ resource "honeycombio_board" "test" {
 			},
 			// finally remove the graph settings and an ensure we're back to the defaults
 			{
-				// skipped due to bug: https://github.com/honeycombio/terraform-provider-honeycombio/issues/399
```
jharley, line 155: Fixed this bug in the process 🥳

docs/resources/board.md
```diff
@@ -104,7 +104,7 @@ The following arguments are supported:
 * `column_layout` - (Optional) the number of columns to layout on the board, either `multi` (the default) or `single`. Only `visual` style boards (see below) have a column layout.
 * `style` - (Optional) Deprecated: All Boards are now displayed as `visual` style. How the board should be displayed in the UI, either `visual` (the default) or `list`.
 * `query` - (Optional) Zero or more configurations blocks (described below) with the queries of the board.
-* `slo` - (Optional) Up to six configuration blocks (described below) to place SLOs on the board.
+* `slo` - (Optional) Up to twenty-four (24) configuration blocks (described below) to place SLOs on the board.
```
jharley, line 107: doc wasn't updated when the limit was increased

internal/provider/board_resource.go
```diff
@@ -0,0 +1,645 @@
...
+							Description: "How the query should be displayed within the Board.",
+							Validators: []validator.String{
+								stringvalidator.OneOf(helper.AsStringSlice(client.BoardQueryStyles())...),
+							},
+						},
+					},
+					Blocks: map[string]schema.Block{
+						"graph_settings": schema.ListNestedBlock{
+							Description: `Manages the settings for this query's graph on the board.
+See [Graph Settings](https://docs.honeycomb.io/working-with-your-data/graph-settings/) in the documentation for more information.`,
```
cewkrupa, line 182: We should update this URL too

internal/provider/board_resource.go
```diff
@@ -0,0 +1,645 @@
...
+				return
+			}
+
+			// GraphSettings are a special case as NestedBlockObjects cannot have a default value
+			// and we need to differentiate between a Null value and a Zero value
+			// to avoid inconsistent state issues
+			if configuredQueries[i].GraphSettings.IsNull() {
+				// if we've not configured a value, we'll keep the Null
+				queryValue["graph_settings"] = configuredQueries[i].GraphSettings
+			}
```
cewkrupa, line 302: Nice, thanks for fixing this bug!

internal/provider/board_resource_test.go
```diff
@@ -326,6 +320,37 @@ resource "honeycombio_board" "test" {
...
+			{
+				ExternalProviders: map[string]resource.ExternalProvider{
+					"honeycombio": {
+						VersionConstraint: "0.32.0",
+						Source:            "honeycombio/honeycombio",
+					},
+				},
+				Config: config,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckBoardExists(t, "honeycombio_board.test"),
```
elimt, line 343: Should we validate the attributes with the migration?

jharley: Not necessary if I had put `PlanOnly: true` which I hadn't until just now. Nice catch!

Doing a "plan only" test will mean the test errors if there is a diff which effectively means all of the fields are tested and no changes are needed from what is written to state

jharley: discovered I had set the wrong default value for `column_layout` in the process 🙏🏻

cewkrupa (approved):

internal/provider/board_resource.go
jharley: nice catch!

PR #660, opened 2025-04-14
feat(d/environment): add detail_filter to simplify environment lookup

jharley:
Received feedback that looking up environment IDs was a bit awkward.

This allows looking up a single environment via a filter: the "no match" behaviour may feel like a breaking change but when `id` was the only option the data source would fail with 404 anyway, so, the same net-result.

### DIFF
--- a/docs/data-sources/environment.md
+++ b/docs/data-sources/environment.md
@@ -22,8 +22,10 @@ data "honeycombio_environment" "prod" {
 
 ```hcl
 data "honeycombio_environment" "classic" {
-  detail_filter = "name"
-  value         = "Classic"
+  detail_filter {
+    name  = "name"
+    value = "Classic"
+  }
 }
 
 data "honeycombio_environment" "prod" {
--- a/docs/data-sources/environment.md
+++ b/docs/data-sources/environment.md
@@ -1,7 +1,10 @@
 # Data Source: honeycombio_environment
 
 The `honeycombio_environment` data source retrieves the details of a single Environment.
-If you want to retrieve multiple Environments, use the `honeycombio_environments` data source instead.
+
+~> **Warning** Terraform will fail unless exactly one environment is returned by the search.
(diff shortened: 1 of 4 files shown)


jharley:
Example of current awkwardness vs with this change:

### Current

```hcl
data "honeycombio_environments" "ci" {
  detail_filter {
    name  = "name"
    value = "ci"
  }
}

data "honeycombio_environment" "ci" {
  id = data.honeycombio_environments.ci.ids[0]
}

output "ci" {
  value = data.honeycombio_environment.ci.id
}
```

### Improved

```hcl
data "honeycombio_environment” "ci" {
  detail_filter {
    name  = "name"
    value = "ci"
  }
}

output "ci" {
  value = data.honeycombio_environment.ci.id
}
```

docs/data-sources/environment.md
```diff
@@ -15,11 +18,36 @@ data "honeycombio_environment" "prod" {
...
+data "honeycombio_environment" "classic" {
+  detail_filter {
+    name  = "name"
+    value = "Classic"
+  }
+}
+
+data "honeycombio_environment" "prod" {
+  detail_filter {
+    name  = "name"
```
elimt, line 33: What do you think about making this value `key` so we do not have to add more fields to this when we start supporting filtering on other fields.
We could add a check for throwing an error if anything that is not `name` is passed

elimt: oh looking at it again. I guess that is what `name` is trying to do. The `name` key confused me

jharley: Yea, it's the name of the field to filter on. I modelled it after the `filter` block in the aws provider

jharley: The code is standard in our provider and that check is in place: https://github.com/honeycombio/terraform-provider-honeycombio/blob/main/internal/helper/filter/detail_filter.go#L17-L19

elimt (approved):

PR #661, opened 2025-04-17
feat(r/api_key): add computed 'key' field

jharley:
Adds the computed attribute `key` containing the API key formatted for usage based on its type (which, is currently only ingest keys).


internal/provider/api_key_resource.go
```diff
@@ -270,10 +287,12 @@ func (r *apiKeyResource) Delete(ctx context.Context, req resource.DeleteRequest,
 	var detailedErr client.DetailedError
 	if err != nil {
 		if errors.As(err, &detailedErr) {
-			resp.Diagnostics.Append(helper.NewDetailedErrorDiagnostic(
-				"Error Deleting Honeycomb API Key",
-				&detailedErr,
-			))
+			if !detailedErr.IsNotFound() {
```
jharley, line 290: fly by bug fix

elimt: If the error is actually `IsNotFound`, do we return something else?

internal/provider/api_key_resource.go
```diff
@@ -173,6 +180,16 @@ func (r *apiKeyResource) Create(ctx context.Context, req resource.CreateRequest,
 		state.Permissions = types.ListNull(types.ObjectType{AttrTypes: models.APIKeyPermissionsAttrType})
 	}
 
+	switch key.KeyType {
+	case "ingest":
+		state.Key = types.StringValue(key.ID + key.Secret)
```
elimt, line 185: Do we typically combine the `ID` and `Secret` for keys? Do clients typically have to do this? I would have guessed those are separate fields

jharley: For ingest keys you need to concatenate the ID and the Secret in order to use it.

The feedback that spawned this PR is that it felt obtuse to push this task onto the end user in Terraform. This `key` field means they no longer have to (see updated example in docs), while paving the way for any other key type to be supported in a similar fashion.

internal/provider/api_key_resource.go
elimt: yeah that is for non-404. If it's a detailedErr and it's `404`, what do we do? we don't return any error?

jharley: Sorry, I misunderstood.

No error: it will be removed from state which is what we want anyway. Forcing the user to deal with an error (and potentially a failed run) that they can't do anything about seems unnecessary and matches the behaviour elsewhere in the provider

elimt (approved):

PR #662, opened 2025-04-21
feat(r/slos): get all slos per environment

kattelles:

The honeycomb API now supports getting all SLOs in an environment by using `__all__` in place of the dataset.

https://app.asana.com/1/72322038700732/project/1205961717360815/task/1207770326676044 

When a user gets all slos, allow them to pass either a dataset or nothing.


docs/data-sources/slos.md
```diff
@@ -1,5 +1,5 @@
 # Data Source: honeycombio_slos
-The SLOs data source retrieves the SLOs of a dataset, with the option of narrowing the retrieval by providing a `detail_filter`.
+The SLOs data source retrieves the SLOs of a dataset, with the option of narrowing the retrieval by providing a `detail_filter`.  To retrieve all of the SLOs in an environment pass the special parameter `__all__` in place of a dataset.
```
kattelles, line 2: I wonder if it's worth adding a note that you'll need an env api key for this?

kattelles:
> The previous release of the provider removed the need for a user to know the magical incantation `__all__`
> 
> Can you update `d/slos` to use `helper.GetDatasetOrAll` so we're not regressing, please :)

ah yes, I was confused about what we needed to do for this task. It's ready for another review :)

internal/provider/slos_data_source_test.go
```diff
@@ -116,4 +116,19 @@ data "honeycombio_slos" "exact" {
...
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 testAccPreCheck(t),
+		ProtoV5ProviderFactories: testAccProtoV5ProviderFactory,
+		Steps: []resource.TestStep{
+			{
+				Config: `
+data "honeycombio_slos" "all" {}
+`,
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr("data.honeycombio_slos.all", "ids.#", "4"),
```
kattelles, line 129: This test is failing because the number of slos isn't deterministic? 🤔

> slos_data_source_test.go:120: Step 1/1 error: Check failed: Check 1/1 error: data.honeycombio_slos.all: Attribute 'ids.#' expected "4", got "6"

jharley: Correct 🙃 

for the moment it's likely better (read: less flakey) to do a filtered version with a known prefix (i.e. `test.$RANDOM.`) and then match on those.

In the future world where we can pragmatically make an Environment and then make things in it this sort of test case will be easier, but we're not there yet.

docs/data-sources/slos.md
```diff
@@ -31,7 +31,7 @@ data "honeycombio_slos" "foo" {
 
 The following arguments are supported:
 
-* `dataset` - (Required) The dataset to retrieve the SLOs list from
+* `dataset` - (Optional) The dataset to retrieve the SLOs list from. If omitted, all SLOs in the environment will be returned.
```
jharley, line 34: slight adjustment as the filter can be used so it won't necessarily be "all"
```suggestion
* `dataset` - (Optional) The dataset to retrieve the SLOs list from. If omitted, the lookup will be Environment-wide.
```

jharley (approved):
Thanks for this! ✨

ronybzzz (approved):
🙌

PR #667, opened 2025-04-30
feat(d/slo,r/slo): add 'tags' support

jharley:
Our first tag-supported resource: SLOs!

This introduces the tags shape to the provider client and adds tags support to `r/slo` and `d/slo`.


client/tags.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+// Tag represents a key-value pair used for tagging resources.
+type Tag struct {
+	Key   string `json:"key"`
+	Value string `json:"value"`
+}
+
+// Validate checks if the tag key and value are valid according to the defined regex patterns.
+func (t *Tag) Validate() error {
```
jharley, line 29: the provider doesn't strictly use `Validate` but it does use these regex's so I added this to make it a little easier to test them.

client/slo_test.go
```diff
@@ -41,60 +42,88 @@ func TestSLOs(t *testing.T) {
...
 			SLI:              client.SLIRef{Alias: sli.Alias},
+			Tags: []client.Tag{
+				{Key: "color", Value: "blue"},
+				{Key: "team", Value: "a-team"},
+			},
 		}
 		slo, err = c.SLOs.Create(ctx, dataset, data)
-
 		require.NoError(t, err, "unable to create SLO")
-		assert.NotNil(t, slo.ID, "SLO ID is empty")
```
jharley, line 48: this test was written in a hacky-lazy way (my bad), so instead of equating two full objects it's now doing more assertions of the various properties

honeycombio/resource_slo.go
```diff
@@ -86,6 +88,36 @@ the column evaluation should consistently return nil, true, or false, as these a
...
+						honeycombio.TagKeyValidationRegex,
+						"must only contain lowercase letters, and be 1-32 characters long",
+					),
+					validation.MapValueMatch(
+						honeycombio.TagValueValidationRegex,
+						"must begin with a lowercase letter, be between 1-32 characters long, and only contain lowercase alphanumeric characters, -, or /",
+					),
+					func(v any, path cty.Path) diag.Diagnostics {
+						// ensure the number of tags is within the resource limit
+						if len(v.(map[string]any)) > honeycombio.MaxTagsPerResource {
```
elimt, line 114: I am not too sure about having a hardcoded limit here. Could we allow the API to throw resource limit error rather? Prevents us from having to release a new TF and centralizes where we control the resource limit in case we decide to change the limit based on team or other attributes

jharley: This is the eternal: "people hate when things don't get caught during a plan" struggle -- especially in workflows like TFC/HCP Terraform, env0 or spacelift. We have tag key and value lengths (and string lenghts, and limit lengths for many other things) encoded in the provider for this reason.

honeycombio/resource_slo.go
```diff
@@ -86,6 +88,36 @@ the column evaluation should consistently return nil, true, or false, as these a
...
 				ValidateFunc: validation.IntAtLeast(1),
 			},
+			"tags": {
+				Type:        schema.TypeMap,
+				Optional:    true,
+				Description: "A map of tags to assign to the resource.",
+				DiffSuppressFunc: func(_, _, _ string, d *schema.ResourceData) bool {
+					// suppress the diff if the tag maps are equivalent
+					oldMap, newMap := d.GetChange("tags")
+					return reflect.DeepEqual(oldMap.(map[string]any), newMap.(map[string]any))
```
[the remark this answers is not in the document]
elimt: I was testing it out here: https://go.dev/play/p/4BoQaKXdx4A

It does return false if the order is different. Could the order ever be different from the API calls?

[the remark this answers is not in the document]
jharley: Your playground example is a slice of maps (`[]map[string]any`), but these are just maps (`map[string]any`): https://go.dev/play/p/HCzonAUuqj3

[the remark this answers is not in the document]
jharley: The response from the API can be in a different order (it came up during while developing this), but this diff suppress func has handled it.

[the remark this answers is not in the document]
elimt: ohh okay got it. This if fine then

honeycombio/resource_slo.go
elimt: > So while two tag updates on a resource could happening at approximately the same time can happen, the last one will win.


I understand what you mean better now. My concern was about multiple updates happening at the same time to the same resource and if having the limit check here could affect it but you raise a good point that it'll always override the latest change

elimt (approved):

PR #675, opened 2025-05-09
feat(boards): Add board type to board resource and client

roncodingenthusiast:

- We are exposing type in the boards public api endpoint as a prep for flexible boards release

- Update to board client 
- update to board resource 
- updating docs


client/board_test.go
```diff
@@ -81,7 +82,6 @@ func TestBoards(t *testing.T) {
 
 		// copy ID before asserting equality
 		data.ID = b.ID
-		data.Queries[0].QueryID = b.Queries[0].QueryID
```
roncodingenthusiast, line 84: this is not needed because the query ID is part of the request

docs/resources/board.md
```diff
@@ -99,49 +100,51 @@ resource "honeycombio_board" "overview" {
 
 The following arguments are supported:
 
-* `name` - (Required) Name of the board.
```
roncodingenthusiast, line 102: `- *` is redundant here

jharley (approved):

PR #676, opened 2025-05-09
feat(boards): Support panels in board client

roncodingenthusiast:

- Breaking down my massive PR into incremental PRs
- This adds panels to board client for flexible boards support

### DIFF
--- a/client/board.go
+++ b/client/board.go
@@ -44,6 +44,10 @@ type Board struct {
 	// Defaults to "classic".
 	BoardType BoardType `json:"type,omitempty"`
 
+	// Board panels are the individual panels that make up a board. Each panel can
+	// be a query or an SLO panel. The panels are laid out in a grid.
+	Panels []BoardPanel `json:"panels,omitempty"`
+
 	// Name of the board, this is displayed in the Honeycomb UI. This field is
 	// required.
 	Name string `json:"name"`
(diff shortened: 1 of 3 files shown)


client/board_test.go
```diff
@@ -143,3 +143,109 @@ func TestClassicBoards(t *testing.T) {
...
+		TargetPerMillion: 990000,
+		SLI:              client.SLIRef{Alias: sli.Alias},
+	})
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		c.SLOs.Delete(ctx, dataset, slo.ID)
+		c.DerivedColumns.Delete(ctx, dataset, sli.ID)
+	})
+
+	t.Run("Create flexible board", func(t *testing.T) {
```
roncodingenthusiast, line 203: Update flexible boards will come as a Separate test since the backend on `main` branch doesn't support that yet.

elimt (approved):

client/board.go
```diff
@@ -65,6 +69,57 @@ type Board struct {
 	SLOs []string `json:"slos"`
 }
 
+// BoardPanel represents a single panel on a board.
+type BoardPanel struct {
```
jharley, line 73: a general "are you sure?": do you mean `omitempty` in all these places here?

jharley (approved):

client/board.go
roncodingenthusiast: yeah, all coming from PAPI too

PR #677, opened 2025-05-12
feat(boards): Add flexible board resource

roncodingenthusiast:

- Adds terraform support for Flexible boards

- create a flexible board using terraform

### DIFF
--- a/client/board.go
+++ b/client/board.go
@@ -97,7 +97,6 @@ type BoardQueryPanel struct {
 	QueryAnnotationID     string                           `json:"query_annotation_id,omitempty"`
 	VisualizationSettings *BoardQueryVisualizationSettings `json:"visualization_settings,omitempty"`
 	Style                 BoardQueryStyle                  `json:"query_style,omitempty"`
-	Dataset               string                           `json:"dataset,omitempty"`
 }
 
 type BoardQueryVisualizationSettings struct {
--- a/client/board_test.go
+++ b/client/board_test.go
@@ -247,5 +247,4 @@ func TestFlexibleBoards(t *testing.T) {
 
 		assert.Equal(t, *board, *flexibleBoard)
 	})
-
 }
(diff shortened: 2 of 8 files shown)


client/board.go
```diff
@@ -97,7 +97,6 @@ type BoardQueryPanel struct {
 	QueryAnnotationID     string                           `json:"query_annotation_id,omitempty"`
 	VisualizationSettings *BoardQueryVisualizationSettings `json:"visualization_settings,omitempty"`
 	Style                 BoardQueryStyle                  `json:"query_style,omitempty"`
-	Dataset               string                           `json:"dataset,omitempty"`
```
roncodingenthusiast, line 100: a field we decide not to expose via the API

client/board_test.go
```diff
@@ -146,6 +146,7 @@ func TestClassicBoards(t *testing.T) {
 
 func TestFlexibleBoards(t *testing.T) {
 	t.Parallel()
+	t.Skip("Skipping flexible board tests until server is fixed")
```
roncodingenthusiast, line 149: will be removed once server side changes are merged

jharley (commented):
A first pass ahead of the API support (mostly nits).

I am concerned about the DX of the empty `position {}` block as it doesn't feel common or, well, ✨ nice ✨ .

docs/resources/flexible_board.md
```diff
@@ -0,0 +1,130 @@
...
+
+- `type` - (Required) Type of the panel.
+- `position` - (Required) Single configuration block to determine position of the panel. Leave block empty when you don't care about panel placement.
+- `slo_panel` - (Optional) This is only required for `type` slo panels. Single configuration block that contains board slo information.
+- `query_panel` - (Optional) This is only required for `type` query panels. Single configuration block that contains board query information.
+
+Each `position` configuration accepts the following arguments:
+
+- `x_coordinate` - (Optional) The x-axis origin point for placing the panel within the layout.
+- `y_coordinate` - (Optional) The y-axis origin point for placing the panel within the layout.
```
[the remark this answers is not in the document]
roncodingenthusiast: yes, if they are not set like `position: {}` should give us `x_coordinate` of `0` and `y_coordinate` of `0`

jharley (changes requested):
Great job working through this! We're nearly there with a few things to fix before we merge

PR #681, opened 2025-05-21
docs: Deprecate board resource in favour of flexible_board

roncodingenthusiast:

Marks board resource as deprecated


client/board_test.go
```diff
@@ -206,6 +206,20 @@ func TestFlexibleBoards(t *testing.T) {
 			BoardType:   "flexible",
 			Description: "A board with some panels",
 			Panels: []client.BoardPanel{
+				{
```
roncodingenthusiast, line 209: opportunistic test improvement

chelshaw (approved):
LGTM!

PR #682, opened 2025-05-26
chore(r/slo): migrate SLO resource to Plugin Framework

elimt:

- Migrate `honeycombio_slo` to Plugin Framework.

migrate slo from SDKv2 to Plugin Framework. The is no change in functionality. There is also no change in schema

1. Create an SLO with the `honeycombio_slo` with `dataset`
2. Update the SLO resource
3. Update the SLO resource with `datasets` and move `dataset`
4. Delete the SLO resource


internal/provider/slo_resource.go
```diff
@@ -0,0 +1,460 @@
...
+		idValue = types.StringValue(dataset)
+	} else {
+		dsValue = types.StringValue(dataset)
+	}
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, &models.SLOResourceModel{
+		ID:       idValue,
+		Dataset:  dsValue,
+		Datasets: types.SetNull(types.StringType),
+		Tags:     types.MapNull(types.StringType),
```
jharley, line 166: nit: I think these should be `Unknown` as, well, we don't know :)
```suggestion
		Datasets: types.SetUnknown(types.StringType),
		Tags:     types.MapUnknown(types.StringType),
```

internal/provider/slo_resource.go
```diff
@@ -0,0 +1,460 @@
...
+		tags, diags := types.MapValueFrom(ctx, types.StringType, tagMap)
+		resp.Diagnostics.Append(diags...)
+		if resp.Diagnostics.HasError() {
+			return
+		}
+		state.Tags = tags
+	} else {
+		// Create an empty map if there are no tags
+		state.Tags = types.MapNull(types.StringType)
+	}
```
jharley, line 235: nit: this is repeated often, it may be worth pulling it into a helper function especially since we know that Triggers and Boards are coming soon

internal/provider/slo_resource.go
```diff
@@ -0,0 +1,460 @@
...
+		datasetsSet, diags := types.SetValueFrom(ctx, types.StringType, ds)
+		resp.Diagnostics.Append(diags...)
+		if resp.Diagnostics.HasError() {
+			return
+		}
+		state.Datasets = datasetsSet
+	} else {
+		// Create an empty set if there are no dataset slugs
+		state.Datasets = types.SetNull(types.StringType)
+	}
```
jharley, line 219: this is repeated often: might be nice to pull this into a function like `flattenDatasetSlugs`

internal/provider/slo_resource.go
```diff
@@ -0,0 +1,460 @@
...
+	var state models.SLOResourceModel
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	dataset := helper.GetDatasetOrAll(state.Dataset)
+
+	var detailedErr client.DetailedError
+	_, err := r.client.SLOs.Get(ctx, dataset.ValueString(), state.ID.ValueString())
```
jharley, line 403: What benefit is there to doing the GET first here?

internal/provider/slo_resource_test.go
jharley, line 1: nice work keeping these tests the same! It makes it easy to see the coverage we had still works here
 
You are missing a test that checks that we gracefully handle taking on a resource made with the Plugin SDK test case which is [recommended in the documentation](https://developer.hashicorp.com/terraform/plugin/framework/migrating/testing#testing-migration). There's a few examples in this project as well.

jharley (commented):
Great work! A couple of things to tidy up ahead of merging, but this is solid 🥳

internal/provider/slo_resource.go
elimt: So with the plugin framework test, I realized DELETE method was getting called at the end of each tests as part of the cleanup. This was affecting `TestAccHoneycombioSLO_RecreateOnNotFound` which deletes the resource manually.

This like in the TF plugin framework does that:  https://github.com/hashicorp/terraform-plugin-sdk/blob/3a612a24e2a320b2b988cbe08d5a892dded3f02a/helper/resource/testing_new.go#L79

internal/provider/slo_resource_test.go
elimt: Added a test for this

internal/provider/slo_resource.go
elimt: that works too. Works better for the case where between the GET and DELETE, the resource gets deleted.

PR #686, opened 2025-05-28
fix(client): use proper format for Retry-After fallback

jharley:
Matching the soon to be released upstream change, this updates `Retry-After` to be parsed properly if we happen to fallback on using it (which, is unlikely to be honest).


chelshaw (approved):
🕥

PR #687, opened 2025-05-28
feat(r/triggers): add support for env-wide triggers

kattelles:

- Closes #<enter issue here>


example/trigger_env_wide/main.tf
```diff
@@ -0,0 +1,50 @@
...
+  }
+  recipient {
+    type   = "marker"
+    target = "Trigger - slow requests" # name of the marker
+  }
+
+}
+
+
+output "result" {
```
ronybzzz, line 48: i think we can remove this output since its mostly for us to test locally

ronybzzz (approved):
just one nit but no other feedback than the ones provided! giving my stamp ahead of time since mine are non-blocking. nice work!

elimt (approved):

jharley:
👋🏻 the docs didn't get touched in this PR, and I suspect they should have. Can we do that in follow-up please? :)

kattelles:
> 👋🏻 the docs didn't get touched in this PR, and I suspect they should have. Can we do that in follow-up please? :)

ah sorry, I updated `docs/resources/trigger.md` but is that not right?

jharley:
> > 👋🏻 the docs didn't get touched in this PR, and I suspect they should have. Can we do that in follow-up please? :)
> 
> ah sorry, I updated `docs/resources/trigger.md` but is that not right?

@kattelles I am so sorry, either that wasn't in my browser view or I missed it. No, that's perfect 🫠

PR #690, opened 2025-06-03
Fix incorrect "prefer_overlaid_charts" field name in flexible board docs

jasonvigil:
May want to call this out in the migration guide for honeycomb_board -> honeycomb_flexible_board, also.


roncodingenthusiast (approved):
https://github.com/honeycombio/terraform-provider-honeycombio/blob/main/internal/provider/flexible_board_resource.go#L224 good catch

elimt (approved):
Validated with the team responsible for this change

PR #694, opened 2025-06-05
feat(r/flexible_board): add support for tags on Boards

chelshaw:

Adds support for tagging (flexible type) Board resources

update create, read and update methods to support tags on boards.

1. Create a flexible board with tags: 
```
resource "honeycombio_flexible_board" "overview" {
  name        = "Now With Tags"
  description = "tag all the things"
  tags = {
    team = "blue"
    env = "staging"
  }
}
```

2. Update the tags and the state should show the latest updated values

### DIFF
--- a/client/board.go
+++ b/client/board.go
@@ -67,6 +67,8 @@ type Board struct {
 	Queries []BoardQuery `json:"queries"`
 	// A list of SLO IDs to be added to the board
 	SLOs []string `json:"slos"`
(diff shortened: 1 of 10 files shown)


jharley (approved):
Thanks for this! ✨ 

Two minor things:

1. the PR title should start with `feat(r/flexible_board): [...]`
2. it would be nice to have one of the `client` tests updated to use the tags (though, I see I missed that when reviewing the triggers addition)

PR #695, opened 2025-06-05
chore(client): add limit and disable series to query result request

cewkrupa:

[Create Query Result](https://api-docs.honeycomb.io/api/query-data/createqueryresult) operations have several newer fields that this client doesn't know about, such as `limit` or `disable_series`. Used in combination, `limit` allows more than 1000 query results (up to 10k) to be returned in a single query.

This PR updates the Client's `QueryResultRequest` to accept the `limit` and `disable_series` fields.

Follow up improvements will add these fields to the query resource. For now, this is just the client.

### DIFF
--- a/client/query_result.go
+++ b/client/query_result.go
@@ -28,7 +28,7 @@ var _ QueryResults = (*queryResults)(nil)
 
 // QueryResult represents a Honeycomb query result.
 //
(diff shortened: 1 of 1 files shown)


jharley (approved):
✨

PR #709, opened 2025-06-27
fix(r/trigger): updating env-wide trigger causes apply error

### LINKED ISSUE #702: r/trigger: Environment-wide trigger apply time error
Updating an environment-wide trigger can result in the following error during apply:

>Error: Provider produced inconsistent result after apply
When applying changes to honeycombio_trigger.example, provider "provider[\"registry.terraform.io/honeycombio/honeycombio\"]" produced an unexpected new value: .dataset: was null, but now cty.StringVal("__all__").
>
>This is a bug in the provider, which should be reported in the provider's own issue tracker.

jharley:
Updating an env-wide trigger was not handling the `dataset` attribute correctly resulting in an apply-time error.

Also noticed that Trigger's didn't get the same update as SLOs to support importing environment-wide triggers by 'bare ID' (e.g. no need to use `__all__`), so fixed that here and updated the test.

- Closes #702

### DIFF
--- a/internal/provider/trigger_resource.go
+++ b/internal/provider/trigger_resource.go
@@ -664,6 +664,39 @@ func (r *triggerResource) ValidateConfig(ctx context.Context, req resource.Valid
 			)
 		}
 	}
+
+	// ensure baseline_details conditions are met
+	if !data.BaselineDetails.IsNull() && !data.BaselineDetails.IsUnknown() {
+		threshold := expandTriggerThreshold(ctx, data.Threshold, &resp.Diagnostics)
+		if resp.Diagnostics.HasError() {
+			return
+		}
+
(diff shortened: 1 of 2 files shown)


internal/provider/trigger_resource.go
```diff
@@ -489,8 +489,8 @@ func (r *triggerResource) Update(ctx context.Context, req resource.UpdateRequest
 	}
 
 	var state models.TriggerResourceModel
-	state.Dataset = dataset
 	state.ID = types.StringValue(trigger.ID)
+	state.Dataset = plan.Dataset
```
jharley, line 493: this is the bugfix

cewkrupa (approved):

Issue #701, opened 2025-06-17
SLI expression changes or SLO detail updates force SLO re-creation


bixu:
**Versions**

Provider: `0.36.0` and `0.36.1`
Terraform: `1.7.1`

**Steps to reproduce**

1. Create a single-dataset SLI/SLO in Honeycomb Classic
2. Check the SLO ID
3. Update the SLI column expression
4. `apply` 
5. Check SLO ID
6. Repeat the process, but change an SLO parameter instead

You should see that the SLO ID has changed.

**Additional context**

We were able to work around this by reverting to version `0.35.0` of the provider.

jharley:
Fixed in #710

PR #710, opened 2025-07-02
fix(r/slo): SLO update causing resource replacement

elimt:
Updating SLO resource after recreation cause the resource to be completely replaced. This was due to `RequiresReplace` forcing a new update everytime.

- add a new plan modifier that only requires the SLO resource to be replaced when `datasets` is changed
- add `UseStateForUnknown` which copies the prior state value, if not null. This will address the bug where the SLO cannot be found during updates

- https://github.com/honeycombio/terraform-provider-honeycombio/issues/701

Updating SLO resource after recreation cause the resource to be completely replaced. This was due to `RequiresReplace` forcing a new update everytime.

- add a new plan modifier that only requires the SLO resource to be replaced when `datasets` is changed
- add `UseStateForUnknown` which copies the prior state value, if not null. This will address the bug where the SLO cannot be found during updates

1. Create a SLO resource and note the ID
2. Change any field in the SLO and note the ID
3. The ID should now be different

### DIFF
--- a/internal/helper/modifiers/datasets_require_replace.go
+++ b/internal/helper/modifiers/datasets_require_replace.go
@@ -1,0 +1,68 @@
+package modifiers
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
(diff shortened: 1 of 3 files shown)


jharley (approved):
Nice work! ✨

PR #722, opened 2025-07-23
rel: v0.37.1

elimt:
Prep for v0.37.1 release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ENHANCEMENTS:
 
-* feat(r/trigger): Accommodate the "dataset_slug" field in triggers (#720)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+# 0.37.1 (July 23, 2025)
+
+ENHANCEMENTS:
+
+* feat(r/trigger): Accommodate the "dataset_slug" field in triggers (#720)
```
jharley, line 5: This one is only for the client -- not the trigger resource. I called it a bug as it was an oversight (and then fits into the patch more cleanly) but I am happy to call it a feature and still ship it in this patch :)

```suggestion
* fix(client): Accommodate the "dataset_slug" field in triggers (#720)
```

jharley (approved):

PR #731, opened 2025-08-13
feat(features): add features support with column import override

### LINKED ISSUE #552: Make dataset + column creation idempotent to avoid punishing "deploy first" workflows
# Problem Statement

- I typically run my services in an environment, and begin collecting traces, before I have defined any boards or triggers in my Terraform code.
- Honeycomb creates new datasets and columns on the fly
- If I deploy to some environments but not others, the Terraform code that I write to create boards and triggers may become valid to run in some environments, but not others, because it refers to columns or datasets that do not exist
- If, instead,  I try to proactively create columns and datasets using Terraform, then that code be invalid in any environment where I _have_ deployed

I'm forced to literally ship every change to production first, and then create a separate pull request later on that adds observability. This is a waste of my time and creates an observability gap.

# Proposed Solution

For boards and columns, the state is trivial and I would argue that this provider should break ranks with Terraform and provide idempotent/upsert semantics. In other words: if the board or column already exists _and if it is identical to the resource specification_, the provider should simply pretend that create succeeded vs. erroring because the resource already exists.

# Workarounds

One can switch to "observe first" i.e. decide on column and dataset names, ship a Terraform change to all environments, and then work on the service in question. That's alienating however.

jharley:
This introduces a `features` block to the provider configuration allowing 'feature toggles' to be provided to resources as we see fit. This design is heavily influenced by the Azure Provider.

The first feature toggle is included, enabling the ability for the columns resource to automatically import a column during creation on conflict (which, was the behaviour prior to v0.12.0 of this provider).

The features block is part of the provider configuration and is designed to be extended per resource as we see fit in the future.

To enable the "import on conflict" behavior for the columns resource you would initialize your provider like this:

```hcl
provider "honeycombio" {
  features {
    column {
      import_on_conflict = true
    }
  }
}
```

This partially resolves #552, and I intend to follow on with an `r/dataset` feature toggle for the same.

Kudos to @elimt for spiking out much of this -- I've listed him as a co-author.

### DIFF
--- a/internal/features/README.md
+++ b/internal/features/README.md
@@ -1,0 +1,140 @@
+# Features Package
+
+The features package is a simple way of providing resource-level "feature toggles" so users of the Honeycomb Provider can modify behavior.
+
+The package only supports Framework-based resources: if the resource you are looking to add a feature toggle for is still based on the PluginSDK please migrate the resource to the Framework as a first step.
+
+## Adding a new Feature
+
+To add new features, follow these steps:
+
+### 1. Define Features Types
+
+In `features.go`, add your new features types:
+
+```go
+// Add to Features struct
+type Features struct {
+    Column  FeaturesColumn
+    Dataset FeaturesDataset  // New feature category
+}
+
+// Add new features struct
+type FeaturesDataset struct {
+    AutoRetention bool
+    EnableCaching bool
+}
+
+// Add Framework model struct
+type FeaturesDatasetModel struct {
+    AutoRetention types.Bool `tfsdk:"auto_retention"`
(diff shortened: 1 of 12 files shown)


internal/features/README.md
cewkrupa, line 1: These internal docs seem really helpful!

cewkrupa (approved):
This looks great! I love the test coverage and extra documentation.

internal/features/README.md
jharley: That was all @elimt! I just edited them to reflect the differences from his spike work to now

PR #732, opened 2025-08-14
feat(r/dataset): add import_on_conflict feature

### LINKED ISSUE #552: Make dataset + column creation idempotent to avoid punishing "deploy first" workflows

jharley:
Building on #731, this adds a similar `import_on_conflict` feature for the dataset resource:

```hcl
provider "honeycombio" {
  feature {
    dataset {
      import_on_conflict = true
    }
  }
}
```

- Closes #552

### DIFF
--- a/client/dataset_test.go
+++ b/client/dataset_test.go
@@ -125,7 +125,7 @@ func TestDatasets(t *testing.T) {
 
 	t.Run("returns a 404 when getting a non-existent dataset", func(t *testing.T) {
 		var de client.DetailedError
-		_, err := c.Datasets.Get(ctx, test.RandomStringWithPrefix("test.", 30))
+		_, err := c.Datasets.GetByName(ctx, test.RandomStringWithPrefix("test.", 30))
 		require.ErrorAs(t, err, &de)
 		assert.True(t, de.IsNotFound())
 	})


internal/provider/column_resource.go
```diff
@@ -57,7 +57,7 @@ func (r *columnResource) Configure(_ context.Context, req resource.ConfigureRequ
 	r.client = c
 
 	features, err := w.Features()
-	if err != nil || features == nil {
```
jharley, line 60: this was silly: the underlying function returns error if features is nil already and even if it was nil it would panic as it calls `err.Error()` 🙃

PR #733, opened 2025-08-14
chore(build): bump Go to 1.24

jharley:
With Go 1.25.0 being released, we're following the "trail latest by one" advice and upgrading to 1.24.x (latest being 1.24.6)


jharley:
Starting over here with #735 -- will properly disable the misleading context background lint from `usetesting`

PR #736, opened 2025-08-15
chore(docs,dataset): add note about feature toggle to r/dataset docs

jharley:
Adds a note to the `r/dataset` documentation indicating that there is a feature toggle for the conflict behavior similar to what was done for `r/column`.


jharley:
This was missed in #732

PR #742, opened 2025-08-20
fix(client): add missing source field to query annotations

roncodingenthusiast:
- Exposes new `source` field in the query annotations client. https://api-docs.honeycomb.io/api/query-annotations/getqueryannotation#query-annotations/getqueryannotation/t=response&c=200&path=source

### DIFF
--- a/client/query_annotation_test.go
+++ b/client/query_annotation_test.go
@@ -42,6 +42,8 @@ func TestQueryAnnotations(t *testing.T) {
 
 		require.NoError(t, err)
 		data.ID = queryAnnotation.ID
+		// validate that the source is one of the allowed values
+		assert.Contains(t, []client.QueryAnnotationSource{client.QueryAnnotationSourceQuery, client.QueryAnnotationSourceBoard}, queryAnnotation.Source)
 		data.Source = queryAnnotation.Source
 		assert.Equal(t, data, queryAnnotation)
 	})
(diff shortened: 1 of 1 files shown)


client/query_annotation_test.go
```diff
@@ -42,7 +42,7 @@ func TestQueryAnnotations(t *testing.T) {
 
 		require.NoError(t, err)
 		data.ID = queryAnnotation.ID
-		data.Source = "query"
+		data.Source = queryAnnotation.Source
```
seh, line 45:
```suggestion
		data.Source = QueryAnnotationSourceQuery
```

client/query_annotation_test.go
```diff
@@ -72,7 +72,7 @@ func TestQueryAnnotations(t *testing.T) {
...
```
seh, line 75:
```suggestion
		data.Source = QueryAnnotationSourceQuery
```

seh (commented):
It looks like these manifest constant names drifted between the primary source and the associated tests.

client/query_annotation_test.go
seh: Oh, I missed the context of use here. I saw the new manifest constant names, looked at the names changed in the test, and recognized that they were not the same. Now I see (better) what you're doing here.

Your trying to handle the challenge that some of the fields are established by the server at creation time, but then you're looking to compare an entire Go struct value to compare that you wound up creating what you asked for. The `assert.Equal` function doesn't handle comparing a subset of the fields of the "data" and "queryAnnotation" variables' values. What you'd really like to do is confirm that the values in the former are an improper subset of the latter.

I don't see [a function in _structchr/testify/assert_](https://pkg.go.dev/github.com/stretchr/testify@v1.10.0/assert#pkg-index) that meets that need.

jharley: @tanya-coding My understanding is that each `t.Run` should be independent and execute fine on it's own: these (in more or less the entire `client` package -- not unique to query annotations) need to be executed in order as most depend on the previous run's state

PR #744, opened 2025-08-22
chore(r/query_annotation): migrate to Plugin Framework

jharley:
Migrates `r/query_annotation` to the Plugin Framework.


internal/provider/query_annotation_resource.go
```diff
@@ -0,0 +1,202 @@
...
+	r.client = c
+}
+
+func (*queryAnnotationResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Query annotations allow you to add notes and descriptions to specific queries for better documentation and context.",
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				Description: "The unique identifier for this Query Annotation.",
+				Computed:    true,
```
cewkrupa, line 59: q - looks like the ApiKeyResource

https://github.com/honeycombio/terraform-provider-honeycombio/blob/448e063ba019b9f3cccb1f08f7b0387fa92f29ce/internal/provider/api_key_resource.go#L70-L71

specifies `Required: false` and `Optional: false` for `id`. I guess those are zero values so they default to false, but should we include them for clarity?

internal/provider/query_annotation_resource.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+func (r *queryAnnotationResource) Delete(ctx context.Context, req resource.DeleteRequest, resp *resource.DeleteResponse) {
+	var state models.QueryAnnotationResourceModel
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	dataset := state.Dataset.ValueString()
```
cewkrupa, line 195: should this use `helper.GetDatasetOrAll` like the rest of the methods here?

cewkrupa (commented):
I think overall seems good, just two small questions

internal/provider/query_annotation_resource.go
jharley: Sure! clarity is nice

internal/provider/query_annotation_resource.go
jharley: Oops! I missed that. Fixed.

PR #751, opened 2025-08-28
feat(queries): Add support for compare_time_offset_seconds

lazybuggy:

- Terraform support for new field `compare_time_offset_seconds` in the `queries` API 
- API PR is [here](https://github.com/honeycombio/hound/pull/26488)

## Asana ticket

https://app.asana.com/1/72322038700732/project/1209727892923597/task/1209727895622899?focus=true

Create a quiery using compare_time_offset_seconds, you should get an error when using an invalid option

These are the valid options (all should be in seconds & can be negative or positive values):
- same time as query time range,
- 30 minutes
- 1 hour
- 2 hours
- 8 hours
- 24 hours
- 7 days
- 28 days
- 182 days

### DIFF
--- a/docs/data-sources/query_specification.md
+++ b/docs/data-sources/query_specification.md
@@ -28,7 +28,7 @@ data "honeycombio_query_specification" "example" {
   filter {
     column = "app.tenant"
     op     = "="
-    value  = "ThatSpecialTenant" 
+    value  = "ThatSpecialTenant"
   }
 
   filter {
@@ -40,9 +40,9 @@ data "honeycombio_query_specification" "example" {
   filter_combination = "AND"
 
   breakdowns = ["app.tenant"]
-    
+
   time_range = 28800 // in seconds, 8 hours
-  
+
   compare_time_offset_seconds = 86400 // in seconds, compare with data from 1 day ago
 }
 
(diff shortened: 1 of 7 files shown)


internal/provider/query_specification_data_source.go
```diff
@@ -89,6 +89,11 @@ func (d *querySpecDataSource) Schema(_ context.Context, _ datasource.SchemaReque
 				Optional:   true,
 				Validators: []validator.Int64{int64validator.AtLeast(0)},
 			},
+			"compare_time_offset_seconds": schema.Int64Attribute{
+				Description: "The time offset for comparison queries, in seconds. " +
+					"Used to compare current time range data with data from a previous time period.",
+				Optional: true,
```
jharley, line 95: This should have a validator checking the possible values:

```suggestion
				Optional: true,
				Validators: []validator.Int64{int64validator.OneOf(
					1800, 3600, 7200, 14400, 28800, 86400, 604800, 2419200, 15724800,
				)},
```

docs/data-sources/query_specification.md
```diff
@@ -65,6 +67,7 @@ The following arguments are supported:
 * `start_time` - (Optional) The absolute start time of the query in Unix Time (= seconds since epoch).
 * `end_time` - (Optional) The absolute end time of the query in Unix Time (= seconds since epoch).
 * `granularity` - (Optional) The time resolution of the query’s graph, in seconds. Valid values must be in between the query’s time range /10 at maximum, and /1000 at minimum.
+* `compare_time_offset_seconds` - (Optional) The time offset for comparison queries, in seconds. Used to compare current time range data with data from a previous time period.
```
jharley, line 70:
```suggestion
* `compare_time_offset_seconds` - (Optional) The time offset for comparison queries, in seconds. Used to compare current time range data with data from a previous time period. Valid values are `1800`, `3600`, `7200`, `14400`, `28800`, `86400`, `604800`, `2419200`, or `15724800`.
```

internal/models/query.go
```diff
@@ -9,20 +9,21 @@ type QueryResourceModel struct {
...
-	Json              types.String                             `tfsdk:"json"` // Computed JSON query specification output
+	ID                       types.String                             `tfsdk:"id"`
+	FilterCombination        types.String                             `tfsdk:"filter_combination"`
+	Breakdowns               []types.String                           `tfsdk:"breakdowns"`
+	Limit                    types.Int64                              `tfsdk:"limit"`
+	TimeRange                types.Int64                              `tfsdk:"time_range"`
+	StartTime                types.Int64                              `tfsdk:"start_time"`
+	EndTime                  types.Int64                              `tfsdk:"end_time"`
+	Granularity              types.Int64                              `tfsdk:"granularity"`
+	CompareTimeOffsetSeconds types.Int64                              `tfsdk:"compare_time_offset_seconds"`
```
jharley, line 20: I wonder if the attribute should just be named `compare_time_offset` for the resource? It's fine in the API and the client, but the `_seconds` feels unnecessary to me especially given that `granularity` is also in seconds and we don't have the suffix there

jharley (commented):
Looking pretty good! A few suggestions and a question

internal/provider/query_specification_data_source.go
lazybuggy: @jharley I didnt include a validator here because a possible value is also whatever the queries time range is (say the user set the time range to `10`, then `10` is also a possible value) so I figured we'd let the API handle the validation logic - is that the right way to go about it 🤔

lazybuggy: 👀 ty for that insight, let me bring that validator logic here
(do we have a way to tell what requests come from terraform? i'd love to see our usage)

jharley (approved):
✨

PR #754, opened 2025-09-05
fix(d/query_specification): handle validation with dynamic blocks

jharley:

- Closes #<enter issue here>


chelshaw (approved):
🙏

PR #762, opened 2025-09-22
fix(query): Allow wider granularity range

jbalboni:

- Closes #<enter issue here>


jharley (approved):
✨

PR #770, opened 2025-10-08
fix(triggers): Trigger Duration/Frequency error message

MadhuVK:

- Closes https://app.asana.com/1/72322038700732/project/1208320199008221/task/1211524769612893?focus=true

The check itself is correct, but the validation error message is incorrect. Fix it up so that it matches our product documentation + expectations.

Reference: https://docs.honeycomb.io/notify/alert/triggers/create/#duration
> Duration can also not be less than the trigger’s frequency.

Test Automation should suffice


chelshaw (approved):
LGTM

PR #776, opened 2025-10-15
Add two trigger with having examples

okmoses:

missing `having` examples

This PR adds a couple examples for how to use the having syntax for triggers, since it's not immediately obvious.

both triggers should apply, assuming a dataset w/the relevant columns. I applied them both locally to confirm.

### DIFF
--- a/docs/resources/trigger.md
+++ b/docs/resources/trigger.md
@@ -54,7 +54,8 @@ resource "honeycombio_trigger" "example" {
 }
 ```
 
-### Example with PagerDuty Recipient and Severity
+### Trigger with PagerDuty Recipient and Severity
+
 ```hcl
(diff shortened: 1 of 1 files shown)


example/trigger_with_having/main.tf
```diff
@@ -0,0 +1,69 @@
+terraform {
```
elimt: Could you please add the example here: https://github.com/honeycombio/terraform-provider-honeycombio/blob/main/docs/resources/trigger.md

This is what shows up on the TF Registry: https://registry.terraform.io/providers/honeycombio/honeycombio/latest/docs/resources/trigger

okmoses: 🤦 oh i looked for that and missed it

PR #789, opened 2025-12-03
docs: migrate docs generation to tfplugindocs with templates

elimt:
Migrate provider documentation to use `tfplugindocs` for automated generation. This enables schema documentation to stay in sync with code changes.

- Use custom `.md.tmpl` templates to control documentation structure and preserve important context (deprecation warnings, usage notes, helpful tables)
- Extract HCL examples into standalone `.tf` files under `examples/`, referenced via `{{tffile}}` for easier maintenance
- Enhance `MarkdownDescription` fields in Go schema definitions to include external documentation links
- Add `make docs` and `make docs-check` for docs generation and CI validation

For reviewers, the main thing to check is that there isn't any regression in anything under `docs/`. I validated all the docs but just in case I missed something

## How to
Generate docs from schema
```
make docs
```

Validate docs are up to date with the schema
```
make docs-check
```

**Big changes:**
The schema docs generation moves away from using `Attribute Reference` heading and uses a `Schema` heading

### DIFF
--- a/.github/workflows/validate-examples.yml
+++ b/.github/workflows/validate-examples.yml
@@ -33,4 +33,4 @@ jobs:
         env:
           TF_IN_AUTOMATION: 1
         run: |
-          TF_CLI_CONFIG_FILE="$GITHUB_WORKSPACE/.terraformrc.local" ./scripts/validate-examples example
+          TF_CLI_CONFIG_FILE="$GITHUB_WORKSPACE/.terraformrc.local" ./scripts/validate-examples examples/samples


elimt:
> This is a huge improvement so even if there are a few errors I think it might be better to ask for forgiveness for any minor regression.
> 
> Two comments but both are non-blocking:
> 
> 1. we already have a top-level `example` directory.. and `example` vs `examples` is pretty confusing. I think it might be worth (can be done in follow up) to move the current `example` into `examples/samples` or `examples/community` or some such thing.
> 2. I think you've checked in three scripts that you used to facilitate this work. Will they serve any future purpose or are they just going to be dead code? If they're just dead code, I think it might be worth not checking them in

@jharley 

1. I moved the exising `example` to `examples/samples`
2. That is deadcode that should not have been added. I deleted that

PR #803, opened 2026-01-16
feat(d/query_specification): Add formulas and calculation filters support

jbalboni:
Note: this is ready to be reviewed, but shouldn't be released until we GA query math (expected to be the end of Feb).

- Closes APPO11Y-3668

Add support for formulas and calculation filters in query specifications:
- Add formula blocks that reference calculations by name
- Add filter blocks within calculations for per-calculation filtering
- Add name attribute to calculations for formula references
- Add validation for duplicate names across calculations and formulas
- Add validation to prevent relational fields with formulas/calculation filters
- Add validation to prevent HEATMAP calculations with formulas

Generally, should be able to use the provider to create queries that use formulas and aggregate filters, assuming it is used with a team and dataset that currently supports those features.

Query math docs: https://docs.honeycomb.io/troubleshoot/product-lifecycle/experimental-features/metrics/query-math/


internal/provider/query_specification_data_source.go
```diff
@@ -142,6 +191,22 @@ func (d *querySpecDataSource) Schema(_ context.Context, _ datasource.SchemaReque
...
+				Description: "Zero or more configuration blocks describing formulas that compute values from calculations.",
+				NestedObject: schema.NestedBlockObject{
+					Attributes: map[string]schema.Attribute{
+						"name": schema.StringAttribute{
+							Description: "The name of the formula.",
+							Required:    true,
+						},
+						"expression": schema.StringAttribute{
+							Description:         "An expression that references calculations by name using the calculated field syntax.",
+							MarkdownDescription: "An expression that references calculations by name. Uses the same syntax as [Calculated Field expressions](https://docs.honeycomb.io/reference/derived-column-formula/), but references calculation names instead of column names.",
```
elimt: If it uses the same syntax, could we add a `validation.IsValidCalculatedField()` validator?

internal/provider/query_specification_data_source.go
```diff
@@ -254,6 +320,73 @@ func (d *querySpecDataSource) Read(ctx context.Context, req datasource.ReadReque
 				"",
 			)
 		}
+
+		// Process calculation filters
```
elimt: yh same. we could move this to `ValidateConfig`

elimt (approved):
suggestion to catch empty string

PR #807, opened 2026-02-02
feat(r/triggers): Query Math Support

MadhuVK:

Fixes https://linear.app/honeycombio/issue/APPO11Y-3668

Add support and documentation to Triggers resource for both Metrics and Query Math. This includes:
* Temporal Aggregation and custom granularity for Metrics Triggers
* A single formula (w/ up to 100 statements) for both Metrics and standard Event Triggers

**These are changes we will be shipping with Metrics GA.** 
* I am marking ready for review to get internal Honeycomb feedback.
* But it is not expected to ship until our GA date, and won't do anything meaningful until the internal feature flags are set for non-Beta customers.

Relevant Beta Documentation: https://docs.honeycomb.io/troubleshoot/product-lifecycle/experimental-features/metrics/
* GA documentation is not publicly available yet

Acceptance/Automated Testing:
1. Metrics Triggers (+ enabled the appropriate datasets)
    1. Simple ones, with the familiar syntax.
    1. More complex ones that exercises all available Metrics customizations:
        1. custom granularity (we choose a default normally)
        1. custom temporal aggregation with a calculated field (there's a default temporal aggregate)
        1. Range interval (normally defaults to the granularity)
1. Formula Expressions

Smoke Test:
* Try using some of the examples to plan/apply changes against your local instance.

### DIFF
--- a/.github/workflows/goreleaser.yaml
+++ b/.github/workflows/goreleaser.yaml
@@ -27,7 +27,7 @@ jobs:
           cache: true
       - name: Import GPG key
         id: import_gpg
-        uses: crazy-max/ghaction-import-gpg@v6
+        uses: crazy-max/ghaction-import-gpg@v7
         with:
           gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
           passphrase: ${{ secrets.PASSPHRASE }}


internal/helper/validation/trigger_query_spec.go
```diff
@@ -0,0 +1,193 @@
...
+
+func (v triggerQuerySpecValidator) Description(_ context.Context) string {
+	return "value must be a valid Trigger Query Specification"
+}
+
+func (v triggerQuerySpecValidator) MarkdownDescription(ctx context.Context) string {
+	return v.Description(ctx)
+}
+
+func (v triggerQuerySpecValidator) ValidateString(ctx context.Context, request validator.StringRequest, response *validator.StringResponse) {
```
elimt: I was looking at the validation again. Should this validate duplicate names across calculations and formulas?

internal/helper/validation/trigger_query_spec_test.go
```diff
@@ -160,6 +160,26 @@ func Test_TriggerQuerySpecValidator(t *testing.T) {
...
 		},
+		"invalid duplicate calculation names": {
+			val: types.StringValue(`{
+				"calculations": [
+					{"op": "COUNT", "name": "total"},
+					{"op": "AVG", "column": "duration_ms", "name": "total"}
+				],
+				"formulas": [{"name": "f", "expression": "$total"}]
+			}`),
+			expectError: true,
```
elimt: ~nit: For these tests, can we assert the error returned? There seem to be multiple paths to return an error so it might be beneficial to ensure we're getting the right error

lizthegrey:
@lizbound ran into this last week.

for posterity (and search engines and any friendly LLM training), this issue is the cause of `only one non-having aggregate is allowed`.

PR #835, opened 2026-04-07
feat: Add auto_investigate support for burn alerts and triggers

roncodingenthusiast:
## Summary

This adds support for auto investigations on burn alerts and triggers for teams with Honeycomb Intelligence enabled.

A new `intelligence` block has been added to the provider's `features` configuration. Teams with Intelligence enabled should set `enabled = true` to unlock the `auto_investigate` attribute on burn alert and trigger resources. Teams without Intelligence are unaffected — upgrading to this version produces no diff on existing resources.

When `auto_investigate` is set to `true` without the intelligence feature enabled, the provider returns a clear validation error guiding the user to enable it.

## Sample Configuration

```hcl
provider "honeycombio" {
  features {
    intelligence {
      enabled = true
    }
  }
}

resource "honeycombio_burn_alert" "example" {
  exhaustion_minutes = 60
  auto_investigate   = true

  dataset = "my-dataset"
  slo_id  = honeycombio_slo.example.id

  recipient {
    type   = "email"
    target = "oncall@example.com"
  }
}

resource "honeycombio_trigger" "example" {
  name             = "High error rate"
  dataset          = "my-dataset"
  auto_investigate = true

  query_json = data.honeycombio_query_specification.errors.json

  threshold {
    op    = ">"
    value = 100
  }

  frequency = 900

  recipient {
    type   = "email"
    target = "oncall@example.com"
  }
}
```

## Testing

After building the provider locally (`go build -o terraform-provider-honeycombio`), you can test with:

```bash
# Plan to see the new attribute
terraform plan

# Apply to create/update resources with auto_investigate
terraform apply

# Verify no diff on a clean run
terraform plan  # should show "No changes"
```

To verify the upgrade path (existing resources without `auto_investigate`), run `terraform plan` against an existing state — it should show no changes

### DIFF
--- a/docs/resources/burn_alert.md
+++ b/docs/resources/burn_alert.md
@@ -155,6 +155,7 @@ resource "honeycombio_burn_alert" "example_alert" {
 ### Optional
 
 - `alert_type` (String) The alert type of this Burn Alert.
+- `auto_investigate` (Boolean) Whether to automatically investigate when this Burn Alert fires. Requires Honeycomb Intelligence to be enabled.
 - `budget_rate_decrease_percent` (Number) The percent the budget has decreased over the budget rate window.
 - `budget_rate_window_minutes` (Number) The time period, in minutes, over which a budget rate will be calculated.
 - `dataset` (String) The dataset this Burn Alert is associated with. Will be deprecated in a future release.
--- a/docs/resources/burn_alert.md
+++ b/docs/resources/burn_alert.md
@@ -155,7 +155,7 @@ resource "honeycombio_burn_alert" "example_alert" {
 ### Optional
 
 - `alert_type` (String) The alert type of this Burn Alert.
-- `auto_investigate` (Boolean) Whether to automatically investigate when this Burn Alert fires. Requires Honeycomb Intelligence to be enabled.
+- `auto_investigate` (Boolean) Whether to automatically investigate when this Burn Alert fires. Requires Honeycomb Intelligence to be enabled for your team in the Honeycomb UI and the intelligence feature block to be set in the provider configuration.
 - `budget_rate_decrease_percent` (Number) The percent the budget has decreased over the budget rate window.
 - `budget_rate_window_minutes` (Number) The time period, in minutes, over which a budget rate will be calculated.
 - `dataset` (String) The dataset this Burn Alert is associated with. Will be deprecated in a future release.
--- a/internal/provider/burn_alert_resource.go
+++ b/internal/provider/burn_alert_resource.go
@@ -14,6 +14,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/booldefault"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringdefault"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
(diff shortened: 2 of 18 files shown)


internal/provider/burn_alert_resource.go
```diff
@@ -244,13 +260,26 @@ func (r *burnAlertResource) Create(ctx context.Context, req resource.CreateReque
...
+	}
+
 	// Get attributes from config and construct the create request
 	createRequest := &client.BurnAlert{
 		AlertType:   client.BurnAlertAlertType(plan.AlertType.ValueString()),
 		Recipients:  expandNotificationRecipients(ctx, plan.Recipients, &resp.Diagnostics),
 		SLO:         client.SLORef{ID: plan.SLOID.ValueString()},
 		Description: plan.Description.ValueString(),
 	}
+	if r.feature.Enabled {
```
Minivera, line 280: Question: Should we check if the planned `AutoInvestigate` is null here to avoid setting the pointer to a possibly default value bool (instead of nil if not set)?

Minivera (approved):
Looking good on my end. Only one question about some of the logic.

docs/resources/burn_alert.md
```diff
@@ -155,6 +155,7 @@ resource "honeycombio_burn_alert" "example_alert" {
 ### Optional
 
 - `alert_type` (String) The alert type of this Burn Alert.
+- `auto_investigate` (Boolean) Whether to automatically investigate when this Burn Alert fires. Requires Honeycomb Intelligence to be enabled.
```
elimt: When you say `Honeycomb Intelligence` here, do you mean the `intelligence` feature block or do you actually mean `Honeycomb Intelligence`

internal/provider/burn_alert_resource.go
```diff
@@ -413,6 +467,12 @@ func (r *burnAlertResource) Update(ctx context.Context, req resource.UpdateReque
 	state.SLOID = types.StringValue(burnAlert.SLO.ID)
 	state.Description = types.StringValue(burnAlert.Description)
 
+	if burnAlert.AutoInvestigate != nil {
+		state.AutoInvestigate = types.BoolValue(*burnAlert.AutoInvestigate)
+	} else {
+		state.AutoInvestigate = plan.AutoInvestigate
+	}
```
MadhuVK, line 473: nit(non-blocking) - seems like this can go below the comment on new line 476, since it applies here too.

MadhuVK (approved):
Burn Alert / Trigger model/resource changes look good to me.

martin308 (approved):

abhipabbi (approved):
lgtm, was an interesting read!

internal/provider/burn_alert_resource.go
roncodingenthusiast: good question, because this attribute has a default of false, the plan always guarantees a concrete value. this field is also set to computed to `true` so even if the user omits it, it will still be false.

templates/index.md.tmpl
```diff
@@ -150,3 +154,9 @@ The `column` block supports the following:
 ---
 The `dataset` block supports the following:
 * `import_on_conflict` - (Optional) This changes the creation behavior of the dataset resource to import and update an existing dataset if it already exists, rather than erroring out. Defaults to `false`.
+
+---
+The `intelligence` block supports the following:
+* `enabled` - (Optional) Set to `true` to enable intelligence features such as `auto_investigate` on triggers and burn alerts. Defaults to `false`.
+
+~> **Note** [Honeycomb Intelligence](https://docs.honeycomb.io) must first be enabled for your team in the Honeycomb UI before using this block. The `intelligence` feature block in the provider tells Terraform that your team has Honeycomb Intelligence enabled and unlocks related attributes like `auto_investigate`.
```
elimt: We should use this URL: https://docs.honeycomb.io/security-compliance/honeycomb-intelligence rather

elimt (approved):

PR #841, opened 2026-04-23
fix(client): defer v2 auth validation until first use

kentquirk:

Tests in EU frequently time out. Traced to dozens or even hundreds of validation calls that weren't even being used. The system has unbounded blocking I/O in a constructor, which it usually didn't care about. 

Fix:  Defer v2 client auth validation until first use. `v2.NewClientWithConfig` was making a blocking `AuthInfo` round-trip during construction, which meant every provider `Configure` call (once per test step, many
  times per `terraform plan`/`apply`) paid a network round-trip — even for operations that never touch v2 endpoints.

  In CI this caused `TestAcc_QuerySpecificationDataSource_validationChecks` to intermittently time out against the EU API: 41 `PlanOnly` steps × a subprocess-Configure-per-step × a blocking `AuthInfo` with
  `context.Background()` (no timeout). A single stalled TCP connect was unbounded and consumed the suite's 20-minute budget. Goroutine dump confirmed the hang was inside `v2.NewClientWithConfig → AuthInfo →
  Transport.getConn`.

  ## Changes

  - `client/v2/client.go`: remove eager `AuthInfo` call from `NewClientWithConfig`. Add `authMu`/`authInfo` fields and a private `teamSlug(ctx)` helper that fetches once on first use, caches on success, and
  re-fetches on error. Drop the now-unused unexported `skipInitialization` flag.
  - `client/v2/api_keys.go`, `client/v2/environments.go`: drop the `authinfo` field; each method resolves the team slug lazily via `client.teamSlug(ctx)`.
  - `client/v2/client_test.go`: update tests to match — the "unauthorized" contract test now calls `AuthInfo` explicitly; construction tests no longer need `skipInitialization`.

  ## User-visible behavior change

  Invalid v2 credentials (`HONEYCOMB_KEY_ID` / `HONEYCOMB_KEY_SECRET`) now surface on first use of a v2 resource or data source (`honeycombio_api_key`, `honeycombio_environment`, `honeycombio_environments`) rather
  than at provider-configure time. Error message is unchanged; only the timing differs. Users of only v1 resources with broken v2 creds will see a strict improvement — those errors no longer block unrelated
  operations.

  No HCL schema changes. No Go public API changes.

  ## Test plan

  - [ ] `go build ./...` and `go vet ./...` clean
  - [ ] `go test ./client/v2/ -run TestClient_Config` passes with no network access
  - [ ] Full acceptance suite passes in US
  - [ ] Full acceptance suite passes in EU (the actual target — validate the timeout fix)
  - [ ] Spot-check: `honeycombio_environment` create with a valid key still works
  - [ ] Spot-check: `honeycombio_environment` create with an invalid key returns the auth error at apply time

### DIFF
--- a/client/v2/client.go
+++ b/client/v2/client.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/hashicorp/go-retryablehttp"
@@ -29,13 +30,12 @@ const (
 )
 
 type Config struct {
-	APIKeyID           string
-	APIKeySecret       string
-	BaseURL            string
-	Debug              bool
-	HTTPClient         *http.Client
-	UserAgent          string
-	skipInitialization bool
+	APIKeyID     string
+	APIKeySecret string
+	BaseURL      string
+	Debug        bool
+	HTTPClient   *http.Client
+	UserAgent    string
 }
 
 type Client struct {
@@ -45,9 +45,31 @@ type Client struct {
 
 	http *retryablehttp.Client
 
+	// authInfo is populated on first use by teamSlug. It caches the team slug
+	// needed to construct v2 API paths so we don't do a blocking round-trip
+	// during client construction.
+	authMu   sync.Mutex
+	authInfo *AuthMetadata
+
 	// API handlers here
 	APIKeys      APIKeys
 	Environments Environments
+}
+
+// teamSlug returns the team slug for the configured API key, fetching the
+// auth metadata from the API on first use and caching the result. Errors are
+// not cached so that a transient failure does not poison later calls.
+func (c *Client) teamSlug(ctx context.Context) (string, error) {
+	c.authMu.Lock()
+	defer c.authMu.Unlock()
+	if c.authInfo == nil {
+		info, err := c.AuthInfo(ctx)
+		if err != nil {
+			return "", err
+		}
+		c.authInfo = info
+	}
+	return c.authInfo.Team.Slug, nil
 }
 
 func NewClient() (*Client, error) {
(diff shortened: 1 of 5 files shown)


client/v2/client.go
```diff
@@ -101,7 +123,7 @@ func NewClientWithConfig(config *Config) (*Client, error) {
 		HTTPClient:   config.HTTPClient,
 		RetryWaitMin: 200 * time.Millisecond,
 		RetryWaitMax: time.Minute,
-		RetryMax:     30,
+		RetryMax:     10,
```
kentquirk, line 126: A maximum possible 30-minute retry seems excessive, so I shortened it.

client/v2/client.go
```diff
@@ -45,11 +45,33 @@ type Client struct {
...
 	// API handlers here
 	APIKeys      APIKeys
 	Environments Environments
 }
 
+// teamSlug returns the team slug for the configured API key, fetching the
+// auth metadata from the API on first use and caching the result. Errors are
+// not cached so that a transient failure does not poison later calls.
+func (c *Client) teamSlug(ctx context.Context) (string, error) {
+	c.authMu.Lock()
```
elimt, line 63: This will block on every request, even after the cache has been filled. Could we not use something like an atomic pointer here instead? So something like:

```
if info := c.authInfo.Load(); info != nil {
		return info.Team.Slug, nil
	}
	v, err, _ := c.authFetch.Do("auth", func() (any, error) {
		if info := c.authInfo.Load(); info != nil {
			return info, nil
		}
		info, err := c.AuthInfo(ctx)
		if err != nil {
			return nil, err
		}
		c.authInfo.Store(info)
		return info, nil
	})
	if err != nil {
		return "", err
	}
	return v.(*AuthMetadata).Team.Slug, nil
```

So only the first call hits the cold path

kentquirk: The code in this PR blocks only for the duration it takes to check a pointer for nil-ness. Any Load() routine would also have to lock as well, and this code is also testing the pointer.

Am I misunderstanding?

elimt: I was more concerned about quickly returning the cache result with Load(). I believe it allows multiple goroutine calls simultaneously vs the Lock, which is down to 1 goroutine at a time.
I am not sure if that has any impact on the changes or optimization

elimt (approved):

PR #846, opened 2026-04-24
feat(d/query_specification): Expanded calculation support

MadhuVK:

Updates the query specification data source to allow the following:
* COUNT w/ a column
* COUNT_DATAPOINTS w/ and w/o a column
* HISTOGRAM_COUNT w/ a column

Fixes: https://linear.app/honeycombio/issue/APPO11Y-4275

Honeycomb metrics support for improved timeseries aggregation included two new calculations (also known as aggregates). These have been added to the API, but were not usable in the provider unless the query was manually constructed with JSON. 

* The changes here will allow us to validate the new calculations during planning without throwing an error.
* Metrics datasets currently require a calculation and rejects an empty calculation block in either spec or JSON form.
  * UI-side also prefers the COUNT_DATAPOINTS aggregate since COUNT is less meaningful. 
  * Unfortunately, it's a bit difficult to check that an aggregate is applied for metrics at plan time since the query data sources don't have dataset information.
  * Open to suggestions here, but I'm mostly deferring this as an apply-time problem since its a bit niche, and mentioned in the docs that calculations are strongly encouraged.
* The COUNT(...) inclusion is mostly a drive-by; I am happy to leave it separate if we prefer; but this has been out for a while now, and I was a bit surprised to find that we didn't support this.

I've added unit tests and acceptance tests.
* Similar to the metrics acceptance tests for triggers, this cannot run in CI today because we cannot appropriately clean up metrics datasets.
* I've verified that they run locally.

```
λ gotestsum --debug --format=testname -- -count=10 -race -parallel=8 ./... -run '(?i)metric' | grep -v EMPTY
exec: [go test -json -count=10 -race -parallel=8 ./... -run (?i)metric]
go test pid: 23669
PASS internal/provider.TestAcc_QueryResourceWith_Metrics (7.57s)
PASS internal/provider.TestAcc_TriggerResourceWithMetrics (5.68s)
PASS internal/provider.TestAcc_QueryResourceWith_Metrics (7.29s)
PASS internal/provider.TestAcc_TriggerResourceWithMetrics (5.36s)
...
```

I've also added examples on various other integrations with this specification with our other entities, like boards, triggers, etc, that we can use for verification.

### DIFF
--- a/client/query_spec.go
+++ b/client/query_spec.go
@@ -123,7 +123,7 @@ func (qs *QuerySpec) EquivalentTo(other QuerySpec) bool {
 			// the dataset, otherwise we'll incorrectly assume that empty and COUNT are
 			// the same for metrics (which doesn't have a default).
 			//
-			// We're ignoring this problem for now since COUNT is strongly discouraged
+			// We're ignoring this problem for now since COUNT is disallowed
 			// for use by metrics queries, and often isn't what users actually want.
 			// See https://docs.honeycomb.io/investigate/query/examples-metrics#common-select-operations
 			return c.Op == CalculationOpCount
--- a/client/query_spec.go
+++ b/client/query_spec.go
@@ -31,8 +31,9 @@ type QuerySpec struct {
 	// API. This value should not be set when creating or updating queries.
 	ID *string `json:"id,omitempty"`
 
-	// The calculations to return as a time series and summary table. If no
-	// calculations are provided, COUNT is applied.
+	// The calculations to return as a time series and summary table. For
+	// non-metrics datasets, COUNT is applied if no calculations are provided.
+	// Metrics datasets require at least one explicit calculation.
 	Calculations []CalculationSpec `json:"calculations,omitempty"`
 	// CalculatedFields are temporary Calculated Fields that are
 	// created for the query.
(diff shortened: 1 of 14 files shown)


client/query_spec.go
```diff
@@ -31,8 +31,9 @@ type QuerySpec struct {
 	// API. This value should not be set when creating or updating queries.
 	ID *string `json:"id,omitempty"`
 
-	// The calculations to return as a time series and summary table. If no
-	// calculations are provided, COUNT is applied.
+	// The calculations to return as a time series and summary table. For
+	// non-metrics datasets, COUNT is applied if no calculations are provided.
+	// Metrics datasets require at least one explicit calculation.
```
lazybuggy, line 36: `COUNT_DATAPOINTS` is auto applied for Metrics if no agg is provided (not sure if the API does this though 🤔 )

lazybuggy (approved):
left 1 comment (not sure if blocking depending on answer) but everything lgtm!

client/query_spec.go
MadhuVK: great observation: the API does _not_ do this, despite the UI affordance.
* I actually had this originally to default to count_datapoints as much as possible.
* And then discovered that the tests failed with an error (see the metrics tests failure cases I added which start with empty blocks).
* https://github.com/honeycombio/terraform-provider-honeycombio/pull/846/commits/881d13ec1486652e1f48d21aebc01e75659c7b80

internal/provider/query_specification_data_source.go
```diff
@@ -416,7 +420,7 @@ func (d *querySpecDataSource) Read(ctx context.Context, req datasource.ReadReque
 
 		calculations = append(calculations, calculation)
 	}
-	// 'COUNT' is the default calculation and will be returned by the API if
+
 	// none have been provided. As this can potentially cause an infinite diff
```
jairo-mendoza: nit: this comment looks to be cut off / hard to understand after the recent change

client/query_spec_test.go
```diff
@@ -117,6 +115,36 @@ func TestQuerySpec_EquivalentTo(t *testing.T) {
...
+				Granularity: client.ToPtr(0),
+				Breakdowns:  []string{},
+				Orders:      []client.OrderSpec{},
+				Limit:       client.ToPtr(client.DefaultQueryLimit),
+			},
+			client.QuerySpec{},
+			false,
+		},
+		{
+			"default w/ column is not equivalent to empty",
```
jairo-mendoza: nit: both tests have the same name, could we make the second more specific to `COUNT_DATAPOINTS`?

jairo-mendoza (approved):

client/query_spec_test.go
MadhuVK: great catch - I intended to test that COUNT_DATAPOINTS doesn't trigger the default equivalence, and forgot to update that. I'll adjust. Appreciate your eagle eyes here 🙇

MadhuVK: Done in https://github.com/honeycombio/terraform-provider-honeycombio/pull/846/commits/60687e09027fad9e77cc4b5942d231d00551527c

internal/provider/query_specification_data_source.go
MadhuVK: done in https://github.com/honeycombio/terraform-provider-honeycombio/pull/846/commits/60687e09027fad9e77cc4b5942d231d00551527c
* Mostly restored with the metrics caveat - let me know if things still seem unclear, and thanks for catching my bad deletion.

chelshaw (approved):
LGTM, thanks!

PR #854, opened 2026-05-13
Always treat query filter values as strings for string ops

### LINKED ISSUE #853: String query filter values converted to integers
I'm encountering the opposite of https://github.com/honeycombio/terraform-provider-honeycombio/issues/27.

I'm trying to express [in terraform](https://registry.terraform.io/providers/honeycombio/honeycombio/latest/docs/data-sources/query_specification#nested-schema-for-filter) a filter for strings starting with a certain digit. However, the HC terraform provider seems to interpret the value in a filter block like
```
{
    column = <col>
    op = "starts-with",
    value = "4"
}
```
as the **integer** 4, not the single-character string "4". The resulting query doesn't work because the column is a string type.

When the provider attempts to [determine the concrete type of the HCL `value` string](https://github.com/honeycombio/terraform-provider-honeycombio/blob/main/internal/provider/query_specification_data_source.go#L411), it interprets the `value = "4"` as the integer 4 because `strconv.ParseInt()` works.

**Versions**
0.45.0

**Steps to reproduce**

1. Configure a query specification using a filter block like above.
2. Observe that the query appearing on Honeycomb regards the prefix to match as an integer, not a string.

**Additional context**

Would it work to always treat values for the filter op "starts-with" as strings?

cjao:
HCL filter string values were being interpreted as numeric API values even for filter ops only defined for strings:

- "starts-with"
- "does-not-start-with"
- "ends-with"
- "does-not-end-with"
- "contains"
- "does-not-contain"

For example, `op = 'starts-with', value = '4'` in HCL becomes {"op": "starts-with", "value": 4} instead of {"op": "starts-with", "value": "4"} in the Honeycomb API request body.

When converting HCL filter values to API filter values, leave the values as strings for such filter operations.

Closes #853


kattelles:
sorry for the delayed review!

kattelles (approved):

cjao:
Thanks @kattelles ! Any other feedback for this change?

jairo-mendoza (approved):
team growth related change LGTM

cjao:
Thanks @jairo-mendoza . I'm assuming CI will test this out at some point?

kattelles (changes requested):
@cjao I'd like to get another member of my team to sign off before merging. I'll try to get some eyes on it today. Thank you!

cjao:
I added a test in query_specification_data_source_test.go modeled after a similar test involving a numerical filter, but don't really have an existing test pattern to follow in board_view_resource_test.go. Any suggestions?

brookesargent (approved):

kattelles (approved):
@cjao thank you for your patience and your contribution! you are good to merge :) 

regarding your CI question: we don't have a great flow for running CI from forks right now but it it will run on merge so we'll be able to fix forward if needed.

cjao:
Thanks for your reviews. I think someone from Honeycomb will need to press the merge button as I don't have that permission.

PR #856, opened 2026-05-20
fix: make it possible to delete a dataset [SCALE-588]

kentquirk:

When attempting to delete a dataset that was created with Terraform, the provider attempts to delete the columns in the dataset. But some of those columns are used in dataset definitions, which causes them to fail. But it's not possible to get them out of the definitions, so it becomes impossible to delete the dataset at all with Terraform (it does work if you do it manually).

This PR ignores deletion errors for columns with dataset definitions; Honeycomb still allows deletion of the dataset in this instance, so this allows the intended behavior to work.

In the case where someone attempts to delete only a column, and it's a column used in a dataset definition, the deletion will appear to succeed but the column will still exist. This is now noted in the documentation.

* Conditionally ignore dataset definition errors
* Add a test to verify it works and fails appropriately
* Add a note to docs

* Create a dataset with the provider
* Send it data with duration_ms
* Delete the dataset -- it should succeed anyway

### DIFF
--- a/internal/provider/column_resource.go
+++ b/internal/provider/column_resource.go
@@ -233,6 +233,14 @@ func (r *columnResource) Delete(ctx context.Context, req resource.DeleteRequest,
 
 	err := r.client.Columns.Delete(ctx, state.Dataset.ValueString(), state.ID.ValueString())
 	if err != nil {
+		var detailedErr client.DetailedError
+		if errors.As(err, &detailedErr) && detailedErr.IsConflict() && strings.Contains(detailedErr.Message, "in use by dataset definition") {
+			// This column is a dataset definition column (e.g. timestamp, duration).
+			// The API blocks individual deletion while the parent dataset exists, but
+			// dataset deletion cleans up columns automatically. If the dataset is also
+			// being destroyed, Terraform will delete it next and the column will be gone.
+			return
+		}
 		resp.Diagnostics.AddError("Error deleting Column", err.Error())
 		return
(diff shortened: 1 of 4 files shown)


internal/provider/column_resource.go
```diff
@@ -233,6 +233,14 @@ func (r *columnResource) Delete(ctx context.Context, req resource.DeleteRequest,
 
 	err := r.client.Columns.Delete(ctx, state.Dataset.ValueString(), state.ID.ValueString())
 	if err != nil {
+		var detailedErr client.DetailedError
+		if errors.As(err, &detailedErr) && detailedErr.IsConflict() && strings.Contains(detailedErr.Message, "in use by dataset definition") {
```
elimt, line 237: we should add a `resp.Diagnostics.AddWarning` here so it does not get swallowed

elimt (commented):
The provider's job is more declarative. It shouldn't be in the business of guessing which 409s to ignore based on matching.

I think this should be dealt with at the API layer, maybe with a `force=true` query param or handle deletes there. The provider will then just call the API

kentquirk:
The problem with "--force=true" (which I did consider) is that it just moves the problem -- in that setup, the provider always needs to specify it even if the user is just trying to delete a column. The problem is that terraform is inadequately expressive (treating every resource completely independently), so I feel like the problem should be fixed in the provider.

elimt (approved):

PR #861, opened 2026-05-28
chore: Update the provider proto version to V6

### LINKED ISSUE #543: Update provider protocol to V6
The provider currently is built against the V5 protocol supporting clients as far back as Terraform 0.12.0.

There is tooling available to translate the V5 resources written in the Plugin SDK to V6 ([link](https://developer.hashicorp.com/terraform/plugin/mux/translating-protocol-version-6-to-5)).

The V6 Protocol is the current standard and moves the minimum supported version of Terraform to 1.0 (released June, 2021), but this will drop support for all previous clients. Telemetry shows that we have about two folks using a pre-1.0 version of Terraform but also with a very old release of the provider.

Minivera:

We're on the old v5 version, which has been replaced by V6 in 2021(!). Terraform provide an update utility for moving from V5 to V6, which is what we're using in this case to avoid any breaking changes.

Closes #543

- Updated the providers using the upgrade tool provided by terraform in `main.go`.
- Updated all the test utilities to generate the provider in the same updated version (through the v5tov6 utility) as `main.go`.

Run the acceptance tests. If they all pass, they we shouldn't have any regressions or breakage due to the update.


main.go
```diff
@@ -25,25 +26,46 @@ func main() {
...
+			upgradedSDKServer, err := tf5to6server.UpgradeServer(
+				context.Background(),
+				honeycombio.Provider(providerVersion).GRPCProvider,
+			)
+			if err != nil {
+				log.Fatal(err)
+			}
+
+			return upgradedSDKServer
+		},
```
elimt, line 45: I don't think we need to call `tf5to6server shim` for Framework version of the provider. Is should natively support it with `NewProtocol6`
```suggestion
providers := []func() tfprotov6.ProviderServer{
		providerserver.NewProtocol6(provider.New(providerVersion)),
		func() tfprotov6.ProviderServer {
			upgradedSDKServer, err := tf5to6server.UpgradeServer(
				ctx,
				honeycombio.Provider(providerVersion).GRPCProvider,
			)
			if err != nil {
				log.Fatal(err)
			}

			return upgradedSDKServer
		},
	}
```

elimt (approved):

brookesargent (approved):

discoduckling (approved):
Growth files look good 👍

jbalboni (approved):

PR #864, opened 2026-06-03
fix: detect calculated_fields changes when diffing trigger/query query_json

### LINKED ISSUE #852: Query specification calculated fields are not updated for triggers
**Versions**
0.49.0

**Steps to reproduce**

1. Start with a trigger that defines a query-local calculated field.
```tf
data "honeycombio_query_specification" "errors" {
  calculated_field {
    name       = "is_failed_http_status"
    expression = "IF(GTE($status, \"400\"), 1, 0)"
  }
  calculation {
    op     = "SUM"
    column = "is_failed_http_status"
  }
  filter {
    column = "event_type"
    op     = "="
    value  = "completions"
  }
  time_range = 900
}
resource "honeycombio_trigger" "errors" {
  name        = "HTTP Errors"
  description = "Managed by Terraform"
  query_json = data.honeycombio_query_specification.errors.json
  frequency  = 300
  threshold {
    op    = ">="
    value = 10
  }
}
```
2. Run terraform apply.
3. Change only the calculated field expression, but keep the same calculated field name.
```tf
calculated_field {
  name       = "is_failed_http_status"
  expression = "IF(AND(GTE($status, \"400\"), NOT(EQUALS($status, \"429\"))), 1, 0)"
}
```
4. Run terraform apply again.
5. Expected: Honeycomb trigger uses the new expression.
6. Actual: Honeycomb can keep using the old calculated field expression because the calculated field name did not change.
7. Workaround: make the calculated field name change when the expression changes, for example:
```tf
calculated_field {
  name       = "is_failed_http_status_v2"
  expression = "IF(AND(GTE($status, \"400\"), NOT(EQUALS($status, \"429\"))), 1, 0)"
}
```
8. The issue is not the query itself. It is that Honeycomb/provider state can treat same-name query-local calculated fields as already existing and not refresh their expression correctly.

krantideep95:
### Problem

When you add, change, or remove an inline `calculated_field` in the `query_json` of a `honeycombio_trigger` or `honeycombio_query` that already exists, `terraform plan` reports no change and the update is skipped. The same `calculated_field` is applied when the resource is first created, and the API evaluates it when the trigger runs.

This is a problem when moving a trigger off a persistent derived column to an inline `calculated_field`: the derived column is destroyed, but the stored trigger query still references it, and the plan shows nothing.

### Cause

`query_json` diffs are suppressed by the `EquivalentQuerySpec` plan modifier, which delegates to `client.QuerySpec.EquivalentTo`. That method compared every query field except `CalculatedFields`, so two specs differing only by `calculated_fields` were treated as equivalent.

### Fix

Compare `CalculatedFields` in `EquivalentTo` the same way as `Formulas`: order-sensitive, with an empty slice treated as equivalent to nil. `honeycombio_query` uses the same comparison, so it is fixed too.

### Tests & docs

- Unit cases in `client/query_spec_test.go` covering equal / nil-vs-empty / added / removed / changed-expression / changed-name / reordered.
- Acceptance test `TestAcc_TriggerResourceWithCalculatedField` exercising a calculated-field-only trigger update (calculation and threshold held constant).
- Doc note now lists `calculated_field` as a supported Standard trigger-query field.

Fixes #852

### DIFF
--- a/client/query_spec.go
+++ b/client/query_spec.go
@@ -135,10 +135,8 @@ func (qs *QuerySpec) EquivalentTo(other QuerySpec) bool {
 		}
 	}
 
-	if !reflect.DeepEqual(qs.CalculatedFields, other.CalculatedFields) &&
-		// an empty CalculatedFields is equivalent to a nil CalculatedFields
-		((qs.CalculatedFields != nil || len(other.CalculatedFields) != 0) &&
-			(len(qs.CalculatedFields) != 0 || other.CalculatedFields != nil)) {
+	// the exact order of calculated fields does not matter, but their equivalence does
+	if !Equivalent(qs.CalculatedFields, other.CalculatedFields) {
 		return false
 	}
 
--- a/client/query_spec.go
+++ b/client/query_spec.go
@@ -135,6 +135,13 @@ func (qs *QuerySpec) EquivalentTo(other QuerySpec) bool {
 		}
 	}
 
+	if !reflect.DeepEqual(qs.CalculatedFields, other.CalculatedFields) &&
+		// an empty CalculatedFields is equivalent to a nil CalculatedFields
+		((qs.CalculatedFields != nil || len(other.CalculatedFields) != 0) &&
+			(len(qs.CalculatedFields) != 0 || other.CalculatedFields != nil)) {
+		return false
+	}
+
 	// The order of Formulas is important for visualization rendering
 	if !reflect.DeepEqual(qs.Formulas, other.Formulas) &&
 		// an empty Formulas is equivalent to a nil Formulas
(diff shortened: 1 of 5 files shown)


client/query_spec.go
```diff
@@ -135,6 +135,13 @@ func (qs *QuerySpec) EquivalentTo(other QuerySpec) bool {
 		}
 	}
 
+	if !reflect.DeepEqual(qs.CalculatedFields, other.CalculatedFields) &&
```
jbalboni: I think we'd want to use Equivalent here, like we do for Filters below, since the order of calculated fields isn't important like it is for formulas.

jbalboni (commented):
Looks good generally, I have one minor request.

client/query_spec.go
krantideep95: thanks, switched to the order-insensitive Equivalent helper.

jbalboni (approved):
LGTM. Thanks for the fix!

PR #865, opened 2026-06-04
feat(client): pace and absorb API rate limits so runs slow down instead of failing

subos2008:
## Summary

The provider handles rate limiting **reactively**: requests fire as fast as Terraform schedules them, and an `HTTP 429` is retried by `go-retryablehttp` up to a fixed cap. Under load that bursts straight into the server's limit, and once the cap is exhausted the run **fails**. This is acute for teams with many triggers/SLOs/burn alerts: a state refresh issues one read per resource against a single **per-team** budget, and several environments planning concurrently share that budget.

This PR makes the client well-behaved under rate limits in two complementary ways. Both are needed — the first smooths a single run; the second is what actually keeps concurrent runs from failing.

## Background: the server's model

Honeycomb enforces a **fixed-window** rate limit, scoped **per-team, per-resource, per-action** (e.g. reads of triggers are limited independently of reads of SLOs). Each window grants a fixed budget — default 120 reads/min — refilled **in full at the next window boundary**, not trickled back continuously. Every response, including a `429`, advertises `RateLimit: limit=…, remaining=…, reset=…` and `RateLimit-Policy: …;w=…`.

## 1. Proactive pacing (the gate)

Adds `limits.RateLimitingTransport`, an `http.RoundTripper` wired into both the v1 and v2 clients. It tracks `limit`/`remaining`/`reset` per `(method, resource)` bucket from response headers, passes requests through while budget remains, and pauses a resource until its window resets when the budget is spent — mirroring the fixed-window server.

- **Single run:** workloads within the limit are untouched; a large refresh is paced smoothly instead of bursting into a wall of 429s (measured ~2× faster, near-zero 429s).
- A continuous-refill limiter (`golang.org/x/time/rate`) was deliberately avoided: it would over-issue mid-window against a fixed-window server and reintroduce 429s.

## 2. Absorbing 429s (the fix for concurrent runs)

The gate can't help across **separate** `terraform` processes — each runs its own gate, can't see the others, and they collectively overshoot the shared per-team budget. So a 429 is treated as what it is — "come back when the window resets", not a failure: the transport waits out the window and **replays the request** (body buffered, so any verb works) up to `rate_limit_retries` times.

- `RetryHTTPCheck` no longer retries 429 — the transport owns rate limiting; `retryablehttp`'s `RetryMax` governs only transport errors and 5xx.
- Configurable via the provider's **`rate_limit_retries`** attribute (and `HONEYCOMB_RATE_LIMIT_RETRIES`), plumbed through both clients and both the Plugin Framework and SDKv2 provider schemas. **Defaults to `10`**, matching the prior 429 retry budget, so default behaviour is unchanged; raise it to ride out heavier contention without failing (at the cost of slower runs).

The net effect: rate limiting **slows a run down instead of failing it**, even when many runs share a team's budget.

## Notes / limitations

- Limits are **per-team**, so concurrent runs still contend for one budget; `rate_limit_retries` keeps them from failing, but the durable fix for *speed* is a higher team limit and/or less concurrency.
- The cold start of each process issues up to `-parallelism` requests before it has observed the limit; those few may 429 and are absorbed.

## Testing

- Unit tests (virtual clock + fake fixed-window server): no-429-under-load, correct pauses at window boundaries, pass-through within budget, independent per-resource buckets, context cancellation, absorbing a 429 then succeeding, and giving up after `rate_limit_retries`.
- `make lint` (golangci-lint v2.9.0): 0 issues. `go vet` and `go test -race ./client/internal/limits/...` pass. Docs regenerated via `make docs`.

### DIFF
--- a/client/internal/limits/ratelimiter_test.go
+++ b/client/internal/limits/ratelimiter_test.go
@@ -163,7 +163,7 @@ func (s *fakeFixedWindow) RoundTrip(req *http.Request) (*http.Response, error) {
 // newTestTransport wraps base in a RateLimitingTransport driven by a virtual
 // clock that only advances when the gate sleeps, making pacing deterministic.
 func newTestTransport(base http.RoundTripper, now *time.Time) *RateLimitingTransport {
-	tr := NewRateLimitingTransport(base)
+	tr := NewRateLimitingTransport(base, DefaultRateLimitRetries)
 	tr.clock = func() time.Time { return *now }
 	tr.sleep = func(_ context.Context, d time.Duration) error {
 		*now = now.Add(d)
@@ -256,7 +256,7 @@ func TestRateLimitingTransport_ContextCancelledWhileWaiting(t *testing.T) {
 	now := start
 
 	srv := &fakeFixedWindow{limit: 1, window: time.Minute, clock: func() time.Time { return now }}
-	tr := NewRateLimitingTransport(srv)
+	tr := NewRateLimitingTransport(srv, DefaultRateLimitRetries)
 	tr.clock = func() time.Time { return now }
 
 	ctx, cancel := context.WithCancel(context.Background())
@@ -277,4 +277,70 @@ func TestRateLimitingTransport_ContextCancelledWhileWaiting(t *testing.T) {
 	require.NoError(t, err)
 	_, err = tr.RoundTrip(req2)
 	assert.ErrorIs(t, err, context.Canceled)
+}
+
+// roundTripFunc adapts a function to an http.RoundTripper.
+type roundTripFunc func(*http.Request) (*http.Response, error)
+
+func (f roundTripFunc) RoundTrip(r *http.Request) (*http.Response, error) { return f(r) }
+
+func TestRateLimitingTransport_Absorbs429(t *testing.T) {
+	t.Parallel()
+
+	start := time.Date(2026, 1, 1, 0, 0, 0, 0, time.UTC)
+	now := start
+
+	// First call is rate limited; the retry succeeds.
+	calls := 0
+	base := roundTripFunc(func(r *http.Request) (*http.Response, error) {
+		calls++
+		h := make(http.Header)
+		h.Set(HeaderRateLimitPolicy, "120;w=60")
+		if calls == 1 {
+			h.Set(HeaderRateLimit, "limit=120, remaining=0, reset=30")
+			return &http.Response{StatusCode: http.StatusTooManyRequests, Header: h, Body: http.NoBody, Request: r}, nil
+		}
+		h.Set(HeaderRateLimit, "limit=120, remaining=119, reset=60")
+		return &http.Response{StatusCode: http.StatusOK, Header: h, Body: http.NoBody, Request: r}, nil
+	})
+
+	tr := newTestTransport(base, &now)
+
+	req, err := http.NewRequest(http.MethodGet, "https://api.honeycomb.io/1/triggers/x", nil)
+	require.NoError(t, err)
+	resp, err := tr.RoundTrip(req)
+	require.NoError(t, err)
+
+	assert.Equal(t, http.StatusOK, resp.StatusCode, "the 429 should be absorbed and the request retried")
+	assert.Equal(t, 2, calls)
+	assert.GreaterOrEqual(t, now.Sub(start), 30*time.Second, "should have waited out the window before retrying")
+}
+
+func TestRateLimitingTransport_AbsorbRespectsMax(t *testing.T) {
+	t.Parallel()
+
+	start := time.Date(2026, 1, 1, 0, 0, 0, 0, time.UTC)
+	now := start
+
+	// Always rate limited: the transport should give up after maxRetries and
+	// surface the 429 rather than looping forever.
+	calls := 0
+	base := roundTripFunc(func(r *http.Request) (*http.Response, error) {
+		calls++
+		h := make(http.Header)
+		h.Set(HeaderRateLimitPolicy, "120;w=60")
+		h.Set(HeaderRateLimit, "limit=120, remaining=0, reset=1")
+		return &http.Response{StatusCode: http.StatusTooManyRequests, Header: h, Body: http.NoBody, Request: r}, nil
+	})
+
+	tr := newTestTransport(base, &now)
+	tr.maxRetries = 2
+
+	req, err := http.NewRequest(http.MethodGet, "https://api.honeycomb.io/1/triggers/x", nil)
+	require.NoError(t, err)
+	resp, err := tr.RoundTrip(req)
+	require.NoError(t, err)
+
+	assert.Equal(t, http.StatusTooManyRequests, resp.StatusCode, "after maxRetries the 429 should surface")
+	assert.Equal(t, 3, calls, "1 initial attempt + 2 retries")
 }
--- a/client/internal/limits/ratelimiter_test.go
+++ b/client/internal/limits/ratelimiter_test.go
@@ -3,8 +3,10 @@ package limits
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"net/http"
 	"sync"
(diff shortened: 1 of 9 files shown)


client/internal/limits/ratelimiter_test.go
```diff
@@ -0,0 +1,280 @@
+package limits
```
elimt, line 1: Let's add a test that tests concurrent requests. I think none of the tests cover scenarios with parallel requests. So if there are parallel requests, for example:


```
// reorderingTransport delays each response by a small random amount *after* the
// inner transport has produced it. The fake server stamps each response's
// "remaining" count at processing time, so delaying delivery shuffles the order
// in which the gate observes those counts relative to the order they were
// produced. That reproduces the real-world condition — stale, out-of-order
// rate limit observations — that stresses the gate's bookkeeping under
// concurrency, rather than relying on incidental goroutine scheduling.
type reorderingTransport struct {
	inner http.RoundTripper
	max   time.Duration
}

func (r *reorderingTransport) RoundTrip(req *http.Request) (*http.Response, error) {
	resp, err := r.inner.RoundTrip(req)
	if err == nil && r.max > 0 {
		//nolint:gosec // non-cryptographic jitter is fine for a test
		time.Sleep(time.Duration(rand.Int63n(int64(r.max))))
	}
	return resp, err
}

// scaledClock reports a simulated time that advances at a fixed multiple of
// real time. It lets the test use a realistic minute-long window — so the
// integer-second RateLimit/reset headers stay meaningful, exactly as in
// production — while the whole test still completes in milliseconds. It is safe
// for concurrent use: Now derives purely from the (immutable) real start time,
// with no shared mutable state.
type scaledClock struct {
	realStart time.Time
	simStart  time.Time
	scale     int64 // simulated nanoseconds per real nanosecond
}

func newScaledClock(scale int64) *scaledClock {
	return &scaledClock{
		realStart: time.Now(),
		simStart:  time.Date(2026, 1, 1, 0, 0, 0, 0, time.UTC),
		scale:     scale,
	}
}

func (c *scaledClock) Now() time.Time {
	return c.simStart.Add(time.Since(c.realStart) * time.Duration(c.scale))
}

// sleep blocks for the real-time equivalent of a simulated duration.
func (c *scaledClock) sleep(ctx context.Context, sim time.Duration) error {
	return sleepFor(ctx, sim/time.Duration(c.scale))
}

// TestRateLimitingTransport_ConcurrentLoad drives the gate the way Terraform
// does: many goroutines hammering a single bucket at once (mirroring
// -parallelism)
func TestRateLimitingTransport_ConcurrentLoad(t *testing.T) {
	t.Parallel()

	const (
		limit   = 30          // per-window server budget
		window  = time.Minute // realistic window; integer-second headers stay meaningful
		workers = 10          // < limit, mirrors Terraform parallelism below the budget
		jobs    = 150         // > limit, forces pacing across several windows
	)

	clk := newScaledClock(1000) // 1ms real == 1s simulated: a 60s window elapses in 60ms
	srv := &fakeFixedWindow{limit: limit, window: window, clock: clk.Now}
	base := &reorderingTransport{inner: srv, max: 200 * time.Microsecond}
	tr := NewRateLimitingTransport(base)
	tr.clock = clk.Now
	tr.sleep = clk.sleep

	work := make(chan int, jobs)
	for i := range jobs {
		work <- i
	}
	close(work)

	start := clk.Now()

	var wg sync.WaitGroup
	var got200, got429 int64
	for range workers {
		wg.Go(func() {
			for range work {
				req, err := http.NewRequest(http.MethodGet, "https://api.honeycomb.io/1/triggers/__all__", nil)
				if err != nil {
					t.Errorf("building request: %v", err)
					return
				}
				resp, err := tr.RoundTrip(req)
				if err != nil {
					t.Errorf("RoundTrip: %v", err)
					return
				}
				switch resp.StatusCode {
				case http.StatusOK:
					atomic.AddInt64(&got200, 1)
				case http.StatusTooManyRequests:
					atomic.AddInt64(&got429, 1)
				}
			}
		})
	}
	wg.Wait()

	require.Equal(t, int64(jobs), got200+got429, "every job should produce a response")
	assert.Zerof(t, got429,
		"client saw %d HTTP 429(s): the gate admitted more than a window allowed under concurrency", got429)

	srv.mu.Lock()
	rejected := srv.rejected
	srv.mu.Unlock()
	assert.Zerof(t, rejected,
		"server rejected %d request(s): the gate over-issued under concurrency", rejected)
}

```

subos2008: Yeah I was just testing 3x concurrent plans of 130 triggers which is close to the customer use case. I can commit the code if you want. So while this PR seems to work as designed it still needs some tweaking to fix the customers current problem.

subos2008: @elimt retries was decreased from 15 to 10 at some point - any idea what the rationale behind that was? I think in order to fix the customer issue I'd want to increase that value.

Claude's suggestion, which I think is mostly reasonable, is pull the 429 retires logic inside my code. Leave the MaxRetries code to handle 500s/transient network issues. But have 429's retry more times, or even indefinitely.

subos2008: FYI: nothing in the code/commits says why the retry limit was toned down:

⏺ Traced it. Your memory's close but the path was 15 → 30 → 10:

  - #372 introduced RetryMax: 15 (with RetryWaitMax: 10s).
  - #629 ("improve V1 api client retry behaviour", v0.31.0) raised it to 30 and bumped RetryWaitMax to 60s — sensible, since that's the PR that switched to honoring the rate-limit reset (longer waits → more retries
  needed to ride out multiple windows).
  - #841 ("fix(client): defer v2 auth validation until first use", Kent Quirk, 2026-04-23, v0.48.0) dropped it 30 → 10 in both the v1 and v2 clients.

  Was a reason given? No — none. PR #841 is entirely about deferring the v2 auth AuthInfo call to fix EU acceptance-test timeouts. The RetryMax: 30 → 10 change isn't mentioned anywhere — not in the problem statement,
  not in the bulleted "Changes" list (which only covers client.go auth fields, api_keys.go, environments.go, tests), not in a comment. It's an undocumented drive-by edit bundled into an unrelated PR.

  The only inferable motive is thematic — the whole PR is about CI runs timing out, and fewer retries means failing fast instead of spending up to ~30 × 60s ≈ 30 min backing off — but that's me guessing; it's not
  stated.

  Two things worth flagging from this:
  1. That change materially cut rate-limit resilience (30 windows of survival → 10) with no rationale, in v0.48.0 — recent. If your customer upgraded into that, their tolerance for the 429 storms literally more than
  halved.
  2. It reinforces why our 429-absorb approach is the right fix: it stops resilience from depending on a magic retry-cap number that someone can quietly halve in an unrelated PR.
  
@kentquirk do you happen to remember if that change to the MaxRetries on 429's being reduced from 30 to 10 was to fix anything?

subos2008: > Let's add a test that tests concurrent requests. I think none of the tests cover scenarios with parallel requests. So if there are parallel requests, for example:
> 
> ```
> // reorderingTransport delays each response by a small random amount *after* the
...
> }
> ```

There's no way not to over issue under concurrency when the processes can't communicate with each other. But I've added a test for the other two assertions in that test

elimt: Kent is out for a while, but it was because we thought it might be too high. I do not think there was any evidence to back it apart from the fact that it felt too high

elimt: So yes my concern was the number of concurrent TF request for the same bucket & key and how we handle for it. [TF parallelism](https://developer.hashicorp.com/terraform/cli/commands/apply#parallelism-n) basically but with each sending a request for reading Triggers

subos2008: Ok, I've added the absorption of 429 specific retries and maxRetries into this code. Also made it settable by the user as a parameter.

Issue #555, opened 2024-10-07
Request: Ability to create honeycomb_api_key resources of configuration keys.


dfsdevops:
**Is your feature request related to a problem? Please describe.**
Currently I am attempting to create environments via terraform, and then be able to further configure them with boards, queries, and triggers. Manually creating the environments and configuration keys for every new EKS cluster I create in terraform is becoming tiresome because it requires a lot of secret management for input variables into terraform. I already have plenty of working code for configuring the environments, but I still have to manually create them and create API keys, which leads to a lot of toil, especially when testing terraform module changes.

**Describe the solution you'd like**
I would like to be able to create a "configuration" type of the `honeycomb_api_key` resource. Currently the only supported key type is "ingest". I would then be able to use the output as variables for an aliased honeycombio provider and finish the configuration.

**Describe alternatives you've considered**
- Forking a CLI command directly to the API.
- Manually creating every environment.

**Additional context**

NA

jharley:
Thanks for opening this issue, @dfsdevops!

This isn't currently possible via the Key Management API (only Ingest Keys can be managed today) but enabling a way to use a Management API Key to access resources currently only available via Configuration Keys is something we are discussing and designing currently.

elimt:
Creating configuration keys is now possible as of [v0.51.0](https://github.com/honeycombio/terraform-provider-honeycombio/releases/tag/v0.51.0)

PR #868, opened 2026-06-05
feat: Add configuration key support to the provider

Minivera:

While our API client does support creating and managing configuration keys, the provider didn't support that functionality. 

This adds the feature plus tests to validate that it works. The feature updates the existing provider framework API key resource, which means this should be fully backwards compatible.

The added validation rule is there to avoid misconfiguration where configuration keys permissions are used with ingest keys, and vice versa. Unfortunately, there are no available packages to create a custom schema for unit testing, so this needs to be tested as acceptance tests.


martin308 (approved):

elimt:
closes https://github.com/honeycombio/terraform-provider-honeycombio/issues/555

PR #876, opened 2026-06-18
fix(r/burn_alerts): Enforce single recipient requirement

### LINKED ISSUE #874: Resource `honeycombio_burn_alert` Fails To Create When Optional Field Is Not Present
**Versions**
- `"honeycombio/honeycombio"` = `0.51.0`

**Steps to reproduce**

1. When implementing `resource "honeycombio_burn_alert" "burn_alert"` do not provide a recipient
2. Terraform Apply

**Additional context**
We found this bug when toggling the burn rate to NOT post anywhere in our lower environments for testing purposes.  We provide a per-environment recipient to allow us to change Slack or Incident IO recipients based on where its deployed.

```bash
 Error: Error Updating Honeycomb Burn Alert
│ 
│   with ....honeycombio_burn_alert.burn_alert,
│   on modules..... line 16, in resource "honeycombio_burn_alert" "burn_alert":
│   16: resource "honeycombio_burn_alert" "burn_alert" {
│ 
│ missing recipients - cannot be blank
```

```bash
resource "honeycombio_burn_alert" "burn_alert" {
  alert_type                   = "budget_rate"
  budget_rate_window_minutes   = var.slo_config.budget_rate_window_minutes
  budget_rate_decrease_percent = var.slo_config.budget_rate_decrease_percent
  description                  = var.slo_config.slo_description

  slo_id = honeycombio_slo.slo.id

  // one or more recipients
 // dynamic "recipient" {
  //   for_each = coalesce(var.recipient_config, [])
  //   content {
  //    type   = recipient.value.recipient_type
  //    target = recipient.value.recipient_target
  //  }
  //}
}
```

Screenshot from [provider docs](https://registry.terraform.io/providers/honeycombio/honeycombio/latest/docs/resources/burn_alert) 

Image

MadhuVK:

- Closes #874 
- Fixes: https://linear.app/honeycombio/issue/APPO11Y-4716

Currently, both the web frontend and the API require that every burn alert has at least one recipient. The provider did not check this enforcement.

While it would arguably be better if we didn't have this restriction, it's currently entrenched in the product, and I'd like to follow up separately on lifting this restriction. For now, it isn't safe to just remove the restriction since some application logic breaks.

Thus, I'm doing the minimal thing here of enforcing this behavior at plan-time, rather than an apply-time failure.

Note: The docs currently still list this as optional. I think this is a limitation of the docs generation on SetNestedBlock - there's currently no way for me to mark this as required, other than via the validation.

1. Test Automation Coverage
2. Run through the repro in the issue - it should fail at plan-time now.

### DIFF
--- a/docs/resources/burn_alert.md
+++ b/docs/resources/burn_alert.md
@@ -161,7 +161,7 @@ resource "honeycombio_burn_alert" "example_alert" {
 - `dataset` (String) The dataset this Burn Alert is associated with. Will be deprecated in a future release.
 - `description` (String) A description for this Burn Alert.
 - `exhaustion_minutes` (Number) The amount of time, in minutes, remaining before the SLO's error budget will be exhausted and the alert will fire.
-- `recipient` (Block Set) Zero or more recipients to notify when the resource fires. (see [below for nested schema](#nestedblock--recipient))
+- `recipient` (Block Set) One or more recipients to notify when the resource fires. (see [below for nested schema](#nestedblock--recipient))
 
 ### Read-Only
 
--- a/internal/provider/burn_alert_resource.go
+++ b/internal/provider/burn_alert_resource.go
@@ -146,7 +146,8 @@ func (*burnAlertResource) Schema(_ context.Context, _ resource.SchemaRequest, re
 			},
 		},
 		Blocks: map[string]schema.Block{
-			"recipient": notificationRecipientSchema(client.RecipientTypes()),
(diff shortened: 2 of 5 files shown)


david-davidson (approved):
Looks good! Agree on the immediate step of bringing the provider in line with the UI and the API, even if we'd like to update the underling behavior someday.

elimt (approved):

PR #900, opened 2026-08-18
fix: avoid perpetual granularity diff on metrics queries/triggers

### LINKED ISSUE #896: honeycombio_trigger always shows granularity change after switching to metrics
**Versions**
Provider 0.53
Terraform 1.15.7

**Steps to reproduce**

1. Switch to time series metrics
2. Run plan with a trigger resource
3. See changes even though nothing changed 

My resource looks like this:

```terraform
data "honeycombio_query_specification" "stuff" {
  calculation {
    op     = "SUM"
    column = "amazonaws.com/AWS/ApplicationELB/UnHealthyHostCount"
  }

  filter {
    column = "amazonaws.com/AWS/ApplicationELB/UnHealthyHostCount"
    op     = "exists"
  }

  breakdowns = ["Dimensions.LoadBalancer", "Dimensions.TargetGroup"]

  time_range = 300
}

resource "honeycombio_trigger" "stuff" {
  name        = "stuff"

  query_json = data.honeycombio_query_specification.stuff.json
  dataset    = "metrics"

  frequency = 300

  threshold {
    op    = ">="
    value = 1
  }

  recipient {
    id = "recipient id"
  }
}
```

As you can see `granularity` isn't set anywhere.

The resulting plan always looks like this, no matter how many times i apply it:

```hcl
  ~ resource "honeycombio_trigger" "stuff" {
        id               = "abcdef"
        name             = "stuff"
      ~ query_json       = jsonencode(
          ~ {
              - granularity  = 30
                # (4 unchanged attributes hidden)
            }
        )
        tags             = {}
        # (6 unchanged attributes hidden)

        # (2 unchanged blocks hidden)
    }
```

MadhuVK:
## What

Fixes #896: `terraform plan` on a `honeycombio_trigger` (or `honeycombio_query`) targeting a metrics dataset shows a perpetual `granularity` diff when the query spec omits `granularity`, even immediately after `apply`.

## Why

Metrics datasets have the Honeycomb API compute and persist a non-zero default `granularity` when a query spec omits it. `client.QuerySpec.EquivalentTo` — the semantic-equality check backing the `query_json` plan modifier — hardcoded `0` as the assumed default on *both* sides of the comparison (a fix from an earlier issue, #404, for a different case: the Query Builder exporting a literal `0`). So an API-defaulted value like `30` never matched the config's implicit default, and the diff never resolved.

## Fix

In `client/query_spec.go`, only compare `granularity` when the config explicitly specifies a value. When config omits it, defer to whatever value is already in state — the user expressed no preference, so the API's default (0, or a metrics-specific non-zero default) is accepted either way. This preserves the original `#404` behavior (explicit `granularity: 0` from Query Builder still matches state) and still catches genuine changes (config explicitly sets a different value than state).

`honeycombio_query` shares the same plan modifier and `EquivalentTo`, so the fix applies transparently there too.

## Test plan

- [x] `client/query_spec_test.go`: added cases to `TestQuerySpec_EquivalentTo` — omitted config granularity vs. non-zero state value → equivalent; explicit config/state mismatch → not equivalent. Passing.
- [x] `internal/provider/trigger_resource_test.go`: added `TestAcc_TriggerResourceWithMetricsOmittedGranularity`, a regression test that applies a metrics trigger with `granularity` omitted, then re-plans the same config and asserts no diff (`PlanOnly`/`ExpectNonEmptyPlan: false`). Requires live credentials + `HONEYCOMB_METRICS_DATASET` to run; not exercised in this sandbox.
- [x] `go build ./...`, `go vet ./...` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Unreleased
+
+## Bugfixes
+
+- fix: avoid perpetual diff when `granularity` is omitted on metrics queries/triggers (#896)
+
 # 0.51.0 (June 11, 2026)
 
 ## Enhancements
--- a/client/query_spec.go
+++ b/client/query_spec.go
@@ -193,8 +193,11 @@ func (qs *QuerySpec) EquivalentTo(other QuerySpec) bool {
 	if !reflect.DeepEqual(qs.StartTime, other.StartTime) || !reflect.DeepEqual(qs.EndTime, other.EndTime) {
 		return false
 	}
-	// Granularity may be exported out of the Query Builder as '0' when not provided
-	if PtrValueOrDefault(qs.Granularity, 0) != PtrValueOrDefault(other.Granularity, 0) {
+	// Granularity: if the config doesn't specify a granularity, accept whatever value
+	// the API has settled on (e.g. metrics datasets default to a non-zero granularity,
+	// while other datasets default to 0/auto, and the Query Builder may export it as
+	// literal '0' when not provided). Only compare when config sets one explicitly.
+	if other.Granularity != nil && PtrValueOrDefault(qs.Granularity, 0) != *other.Granularity {
 		return false
 	}
 	if !reflect.DeepEqual(qs.CompareTimeOffsetSeconds, other.CompareTimeOffsetSeconds) {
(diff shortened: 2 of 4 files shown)


aburgel:
@MadhuVK any update on this? admittedly, this is minor and there is an easy workaround, but it would be nice to get it fixed.

MadhuVK:
@aburgel thanks for your patience on this one. This fell of the radar a little bit, but we're picking this up again this week to re-examine.

PR #901, opened 2026-08-19
fix: track notification variables in resource state

9999years:

Burn alert live notification variables returned by Honeycomb were discarded during import and refresh, hiding remote drift from Terraform.

Preserve the configured recipient identity while rebuilding notification details from the API so variables can be imported, tracked, and reconciled.

See tests, I think? I can't run the tests locally because they seem to require live Honeycomb API keys (!), so I'm not sure how to validate this change.

(This code is Claude-written so I don't feel super confident in it, some pointers would be appreciated.)


lazybuggy (approved):
lgtm, thankyou for fixing this!

elimt:
@9999years It looks like you have some failing CI tests on `TestAcc_TriggerResource `

PR #904, opened 2026-09-01
feat(client): serve column reads from a cached list

mterhar:
## Problem

Every column read is its own request — `GET /1/columns/{dataset}?key_name=…` — so a plan or refresh over N managed columns issues N requests in a burst against the collection route's rate limit budget.

For one large IaC fleet this is now **the worst endpoint on the API**, and it is accelerating:

| | requests (7d) | 429s | 429 share |
|---|---|---|---|
| week ending Aug 25 | 234,676 | 97,108 | 41% |
| **week ending Sep 1** | **673,505** | **537,234** | **80%** |

That is 2.9× the volume and 5.5× the rate-limited responses week over week, and it is 60% of that fleet's total API traffic. The shape is unambiguous — 99.99% of it is key-name lookups:

| count | method | status | has `key_name` |
|---|---|---|---|
| 537,234 | GET | **429** | yes |
| 133,208 | GET | 200 | yes |
| 2,973 | GET | 404 | yes |
| 49 | POST | 201/404 | no |

Forty-nine writes against 673k reads. Because the client retries 429s, a large share of those reads are retries of the *same* logical lookup.

## Fix

Extend the read cache added for derived columns in #886 to columns. The generic `cache.Cache[T]` does the work already: the first read fetches the dataset's full column list once, concurrent readers share that single fetch, and subsequent reads are answered from memory. **N per-resource reads collapse into 1 list request.**

- Writes (`Create`/`Update`/`Delete`) invalidate the dataset's cached list.
- A key name absent from the cache falls back to the direct `?key_name=` lookup, so 404/drift semantics are unchanged and a failed list degrades to exactly today's behaviour.
- Columns are always dataset-scoped — there is no environment-wide column — so unlike derived columns there is no `InvalidateAll` case.
- `Get` (by ID) is left uncached; it isn't on the hot path.

### Opt-in

This reuses the **existing** `features { client { read_caching = true } }` flag rather than adding a new one. Nothing changes for anyone who hasn't explicitly asked for it, and operators who have already opted in for derived columns get columns on the same decision.

The one behaviour worth calling out, now documented in the provider docs: `last_written_at` is served from the same cache, so it can lag actual write activity by up to the 60s TTL. It is a `Computed` attribute in both the resource and the data source, so this produces no plan churn — but anyone depending on that timestamp being current should leave `read_caching` off.

## Testing

New tests cover the collapse (50 concurrent `GetByKeyName` calls → exactly 1 list request, 0 key-name requests), that caching is **off by default** (3 reads → 3 direct lookups, 0 lists), miss-fallback propagating `IsNotFound()`, invalidation on create/update/delete, and per-dataset isolation including that a write to one dataset doesn't invalidate another. Passes with `-race`. Existing derived-column, cache, and features suites still pass; `go build`, `go vet`, and `gofmt` clean; docs and template kept in sync.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -50,7 +50,7 @@ type Config struct {
 	// With read caching enabled, reads of supported resource types are
 	// served from a short-lived cache of the containing collection, so
 	// large fleets don't issue one API request per resource read.
-	// Currently supported: derived columns.
+	// Currently supported: columns, derived columns.
 	ReadCaching bool
 }
 
@@ -164,7 +164,11 @@ func NewClientWithConfig(config *Config) (*Client, error) {
 	client.Auth = &auth{client: client}
 	client.Boards = &boards{client: client}
 	client.BoardViews = &boardViews{client: client}
-	client.Columns = &columns{client: client}
+	columns := &columns{client: client}
+	if cfg.ReadCaching {
+		columns.cache = cache.New[Column](columnCacheTTL)
+	}
+	client.Columns = columns
 	client.Datasets = &datasets{client: client}
 	client.DatasetDefinitions = &datasetDefinitions{client: client}
 	derivedColumns := &derivedColumns{client: client}
--- a/client/column.go
+++ b/client/column.go
@@ -5,7 +5,15 @@ import (
 	"fmt"
 	"net/url"
 	"time"
+
+	"github.com/honeycombio/terraform-provider-honeycombio/client/internal/cache"
 )
+
+// columnCacheTTL is how long a dataset's cached column list is served
+// before being refetched. Kept short: the cache exists to collapse the
+// burst of per-resource reads within a single Terraform or Pulumi
+// operation, not to avoid refetching across operations.
+const columnCacheTTL = time.Minute
 
 // Columns describe all the columns-related methods that the Honeycomb API
 // supports.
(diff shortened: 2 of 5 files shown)


client/client.go
```diff
@@ -50,7 +50,7 @@ type Config struct {
 	// With read caching enabled, reads of supported resource types are
```
elimt, line 50: Let's update these docs references to say supported on derived columns and columns.
- https://github.com/honeycombio/terraform-provider-honeycombio/blob/72a67a4f45ae77f6c6200d6f7858e90216470f71/internal/features/schema.go#L24
- https://github.com/honeycombio/terraform-provider-honeycombio/blob/72a67a4f45ae77f6c6200d6f7858e90216470f71/internal/features/schema.go#L87

client/column.go
```diff
@@ -88,9 +103,15 @@ func ColumnTypes() []ColumnType {
 }
 
 func (s *columns) List(ctx context.Context, dataset string) ([]Column, error) {
```
elimt, line 105: The cache gives each reader its own copy of the list. But `Column` has two fields that are pointers (`Hidden` and `Type`) so copying the list doesn't copy what those point at. So every reader gets a different list holding the same two pointers.

If anything ever writes through one of them (`*col.Hidden = true`), it edits the cache itself. To not increase the scope of this PR, we could just fix this for columns for now.

We could do something like

```
func cloneColumns(columns []Column) []Column {
	cloned := slices.Clone(columns)
	for i := range cloned {
		if cloned[i].Hidden != nil {
			hidden := *cloned[i].Hidden
			cloned[i].Hidden = &hidden
		}
		if cloned[i].Type != nil {
			columnType := *cloned[i].Type
			cloned[i].Type = &columnType
		}
	}
	return cloned
}
```

and in `List`

```
columns, err := s.cache.Get(ctx, urlEncodeDataset(dataset), fetch)
if err != nil {
	return nil, err
}
return cloneColumns(columns), nil
```

We can fix this properly at the cache in a separate PR

PR #906, opened 2026-09-04
feat(r/trigger): support grouped recipient routing

martin308:
## Summary

The [Triggers API](https://docs.honeycomb.io/api/triggers/) now supports per-group recipient routing. A Trigger whose query groups by one or more columns can route each group to a different recipient, and open one PagerDuty incident per group. This adds provider support for it.

- `alert_type = "on_group_change"` — resolves each group of a grouped query independently; required for any routing below
- `recipient.group_filter` — maps a group by column to the values which route to that recipient; omit it for a catch-all
- `recipient.pagerduty_per_group_incidents` — one incident per group instead of one per Trigger

```hcl
recipient {
  id                            = data.honeycombio_recipient.pd_checkout.id
  group_filter                  = { "service.name" = ["checkout", "cart"] }
  pagerduty_per_group_incidents = true
}

recipient {                                    # catch-all: hears about every group
  id = data.honeycombio_recipient.slack_oncall.id
}
```

These are available to teams with grouped resolution alerts enabled. Without it the API returns a 422, which surfaces as a plan/apply error; there's no provider-side gate, since unlike `auto_investigate` nothing here needs a client-side opt-in.

## The recipient block is forked; burn alerts are untouched

`recipient` is shared with `honeycombio_burn_alert`, but these attributes are Trigger-only, and the Burn Alerts API rejects unknown fields. Rather than put them on the shared block and reject them for burn alerts, the model is forked. `TriggerNotificationRecipientModel` embeds the shared model, so the existing helpers are reused rather than copied.

The acceptance criterion: **`docs/resources/burn_alert.md` is byte-identical after `make docs`**, and its schema and state shape are unchanged.

The cost is two object types that must stay in sync by hand, so there are two guard tests — one that the AttrType maps agree, one that each schema matches its AttrType. The second is the drift that actually bites: it's a runtime failure on Read, not a compile error.

> **Note on `Optional` vs `Optional+Computed`:** both new attributes are `Optional` only, unlike `notification_details.pagerduty_severity` next door. Because this resource writes config back to state after apply, making them Computed would break set-element correlation when the attribute is removed from a config, and make them impossible to remove at all. Happy to expand in review if useful.

## Validation

The rules the API enforces are checked at plan time where the provider can see enough to do so: routing requires `on_group_change`; `group_filter` may only name the query's group by columns; `pagerduty_per_group_incidents` is PagerDuty-only; and only one routing rule is allowed per recipient.

Two are only partially checkable and fall through to the API — a recipient given by `id` has an unknown type at plan time, and a `query_id` Trigger's group by columns aren't visible to the provider.

The duplicate-recipient check is deliberately scoped to recipients which configure routing, so a configuration that listed the same recipient twice before this feature keeps whatever behaviour it had. It can widen later if that turns out to be worth catching.

## Tests

- **Unit:** wire-shape marshalling (guards the embedded client type staying flat), the `reconcileRead` fork, group-filter flatten/expand, and `validateGroupedRecipientRouting` — 11 cases, most pinning that unknown values are tolerated rather than erroring.
- **Acceptance:** `TestAcc_TriggerResource_groupedRecipientRouting` (create, no-op re-plan, clearing the routing) and `..._groupedRecipientRoutingValidation` (each plan-time rule).
- **Client:** `TestTriggers_GroupedRecipientRouting` round-trips routing through the API — the only coverage of what's actually stored, since state comes from config.

The acceptance and client tests need grouped resolution alerts enabled on the team they run against. Green in both US and EU.

### DIFF
--- a/client/trigger.go
+++ b/client/trigger.go
@@ -77,14 +77,35 @@ type Trigger struct {
 	// divisible by 60 and between 60 and 86400 (between 1 minute and 1 day).
 	Frequency int `json:"frequency,omitempty"`
 	// Recipients are notified when the trigger fires.
-	Recipients      []NotificationRecipient `json:"recipients,omitempty"`
-	BaselineDetails *TriggerBaselineDetails `json:"baseline_details,omitempty"`
+	Recipients      []TriggerNotificationRecipient `json:"recipients,omitempty"`
+	BaselineDetails *TriggerBaselineDetails        `json:"baseline_details,omitempty"`
 	// AutoInvestigate enables automatic investigation when this trigger fires.
 	// Requires Honeycomb Intelligence to be enabled on the team.
 	AutoInvestigate *bool `json:"auto_investigate,omitempty"`
 	// Tags are used to categorize triggers. They can be used to filtering triggers
 	// and are useful for grouping triggers together.
 	Tags []Tag `json:"tags"`
+}
+
+// TriggerNotificationRecipient is a Recipient attached to a Trigger. It extends
+// NotificationRecipient with the per-group routing settings, which are accepted by the
+// Triggers API only -- the Burn Alerts API rejects unknown fields.
+//
+// NotificationRecipient is embedded by value and has no MarshalJSON of its own, so its
+// fields are promoted into a flat JSON object. Adding a MarshalJSON to
+// NotificationRecipient would silently hijack marshalling of this type.
+type TriggerNotificationRecipient struct {
+	NotificationRecipient
+
+	// GroupFilter maps a group by column of the Trigger's query to the values which
+	// route to this Recipient. Empty or absent means this Recipient is notified about
+	// every group. Requires the Trigger's AlertType to be
+	// [TriggerAlertTypeOnGroupChange] and its query to have at least one group by.
+	GroupFilter map[string][]string `json:"group_filter,omitempty"`
+	// PDPerGroupIncidents opens and resolves one PagerDuty incident per triggered
+	// group instead of one per Trigger. PagerDuty Recipients only: the API returns
+	// null for every other Recipient type.
+	PDPerGroupIncidents *bool `json:"pagerduty_per_group_incidents,omitempty"`
 }
 
 type TriggerBaselineDetails struct {
@@ -102,7 +123,8 @@ type TriggerThreshold struct {
 // TriggerThresholdOp the operator of the trigger threshold.
 type TriggerThresholdOp string
 
-// TriggerAlertType determines the alert type of a trigger. Valid values are 'on_change' or 'on_true'
+// TriggerAlertType determines the alert type of a trigger.
+// Valid values are 'on_change', 'on_true', or 'on_group_change'
 type TriggerAlertType string
 
 // TriggerEvaluationScheduleType determines the evaluation schedule type of a trigger. Valid values are 'frequency' or 'window'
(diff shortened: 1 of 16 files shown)


client/trigger.go
```diff
@@ -77,8 +77,8 @@ type Trigger struct {
 	// divisible by 60 and between 60 and 86400 (between 1 minute and 1 day).
```
elimt: Let's update these docs references to say supported on derived columns and columns.
- https://github.com/honeycombio/terraform-provider-honeycombio/blob/72a67a4f45ae77f6c6200d6f7858e90216470f71/internal/features/schema.go#L24
- https://github.com/honeycombio/terraform-provider-honeycombio/blob/72a67a4f45ae77f6c6200d6f7858e90216470f71/internal/features/schema.go#L87

internal/provider/trigger_notification_recipients.go
```diff
@@ -0,0 +1,413 @@
...
+					`Per-group recipient routing requires an "alert_type" of "on_group_change". `+
+						"Honeycomb only evaluates per-group state for that alert type, and will "+
+						"otherwise discard the routing configured here.",
+				)
+			}
+		}
+
+		// pagerduty_per_group_incidents is only supported for PagerDuty recipients.
+		// A recipient given by ID has an unknown type, so this can only be checked when
+		// the type was configured explicitly.
```
david-davidson, line 135: If a recipient is given by ID and is misconfigured (e.g. Slack type but `pagerduty_per_group_incidents`), I don't believe we ever fail or surface this, right? When we get back a server response (which would drop the invalid recipient in the case above), we still build `state.Recipients` from the original configuration, not the server's response (`trigger`): https://github.com/honeycombio/terraform-provider-honeycombio/blob/30d9b85cf81670c1f441c3f63784eca7348936af/internal/provider/trigger_resource.go#L352-L373

Of course, per the comment on :372, I don't see a good way around that. Can imagine two mitigating alternatives:
* [lighter] Loop over `trigger.Recipients`, comparing each `PDPerGroupIncidents` against what the config asked for. On mismatch, use `resp.Diagnostics.AddWarning` to log something like:
```
Honeycomb ignored `pagerduty_per_group_incidents` for recipient <id>: it is a slack recipient, and per-group incidents are only supported for PagerDuty. Terraform state will record the value you configured.
```
* [heavier] For each recipient that sets `pagerduty_per_group_incidents`, GET it by ID and check type

Since the latter is extra API traffic (and I don't believe a pattern we've used before), I lean toward just adding the first and seeing if this comes up in practice.

internal/provider/trigger_notification_recipients.go
```diff
@@ -0,0 +1,413 @@
...
+				recipPath.AtName("pagerduty_per_group_incidents"),
+				"Trigger validation error",
+				`"pagerduty_per_group_incidents" is only supported for PagerDuty recipients, `+
+					"but this recipient is of type "+rcpt.Type.ValueString()+".",
+			)
+		}
+
+		// Only one routing rule is allowed per recipient. Scoped to recipients which
+		// actually configure routing so that configurations which repeated a recipient
+		// before this feature existed keep whatever behaviour the API gave them.
```
david-davidson:
```
		// Scoped to recipients which
		// actually configure routing so that configurations which repeated a recipient
		// before this feature existed keep whatever behaviour the API gave them.
```
I don't believe this is actually possible? Per https://github.com/honeycombio/hound/pull/38698/changes#diff-1c73db0c7132bdd12e3ca8b14fdee02587a7ea8f0da04e31ebe6f3f83c60d49fR365-R366, simply matching on recipient ID is enough to reject validation at the API level. At the DB level, I don't believe a trigger can have two rows for the same recipient.

Takeaway: we can remove the `groupFilterSet || perGroupIncidentsSet` scoping, bringing this in line with the API.

david-davidson (commented):
Two comments in the diff - directionally looks good!

david-davidson (approved):
Looks good (with the two minor changes I just pushed)! @martin308 LMK if anything there looks wrong on your end.

PR #910, opened 2026-09-15
feat(notify): [AD] Anomaly support in terraform

lazybuggy:

Add support for Anomaly Detection as a Webhook type. 
Currently behind a feature flag that will be enabled Sept 23

https://linear.app/honeycombio/issue/APPO11Y-5052/ad-wh-add-terraform-support-for-anomaly-webhook-templates

I tested locally
- terraform plan accepted type = "anomaly"
-  terraform apply creates the recipient & the anomaly payload is visible in the Integration Center


lazybuggy:
@hny-review

client/recipient.go
```diff
@@ -90,6 +90,7 @@ type PayloadTemplates struct {
 	Trigger        *PayloadTemplate `json:"trigger,omitempty"`
 	ExhaustionTime *PayloadTemplate `json:"exhaustion_time,omitempty"`
 	BudgetRate     *PayloadTemplate `json:"budget_rate,omitempty"`
+	Anomaly        *PayloadTemplate `json:"anomaly,omitempty"`
```
sunjayBhatia, line 93: to be pedantic and guard against any regressions we could add a unit test in `recipient_test.go` to cover the marshal/unmarshal of the new field, looks like we only specifically test a custom webhook with a trigger recipient payload

sunjayBhatia (approved):
non-blocking: consider adding an example to https://github.com/honeycombio/terraform-provider-honeycombio/blob/main/examples/resources/honeycombio_webhook_recipient/custom_webhook.tf, looks like we only have a trigger example here

docs/resources/webhook_recipient.md
```diff
@@ -100,7 +100,7 @@ Optional:
 
 - `default_value` (String) An optional default value for the variable
 
-When configuring custom webhook payloads, use the `template` block. The `type` attribute can be `trigger`, `exhaustion_time`, or `budget_rate`. Only one template block of each type is allowed on a single recipient.
+When configuring custom webhook payloads, use the `template` block. The `type` attribute can be `trigger`, `exhaustion_time`, `budget_rate`, or `anomaly`. Only one template block of each type is allowed on a single recipient.
```
elimt, line 116: Can we also add an example for configuring this here, as we did for trigger:

- https://github.com/honeycombio/terraform-provider-honeycombio/blob/main/examples/resources/honeycombio_webhook_recipient/resource.tf
- https://github.com/honeycombio/terraform-provider-honeycombio/blob/main/examples/resources/honeycombio_webhook_recipient/custom_webhook.tf

Just makes it easier for folks to figure out how to configure it

elimt (approved):

docs/resources/webhook_recipient.md
lazybuggy: done! thanks @elimt, when does the next TF release get cut / when is it ok to merge this?

elimt: thanks! you are good to merge. When do you need this out by? The next scheduled release is in about 2 weeks but we can cut a release sooner if necessary

lazybuggy: ooh, is it possible to get a release cut sometime this week? i'll reach out on slack

